PR #4, opened 2021-09-21
Feature/osm loaders

Evelyn-M:
This is the first part of the openstreetmap module, which deals with obtaining map data from osm such that it can be used for exposures, etc. in the next step.

The review is hence mainly the following:
* osm_dataloader.py - contains 3 classes with methods (one for obtaining raw data, one for querying this raw data, one for loading data directly from the OSM api).
* tutorial

I have not written tests, yet. Will do this once the functions are "approved" as such.


climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+Adapted from the GitHub repo osm_clipper (https://github.com/ElcoK/osm_clipper)
+Used by OSMRaw().get_data_geofabrik().
+
+Note: A few small countries will be downloaded as a multi-country file, as
+indicated in the comments.
+
+Note: "special" ISO-3 codes - Canary Islands (IC), Asian part of Russia (RUS-A),
+European part of Russia (RUS-E)
+"""
+DICT_GEOFABRIK = {
```
sameberenz, line 57: Instead on initiating this dictionary in the code, you could consider to put the dictionary into the config file ([climada_petals/conf/climada.conf](https://github.com/CLIMADA-project/climada_petals/blob/feature/osm_loaders/climada_petals/conf/climada.conf)) and load it from the `CONFIG` variable, maybe as `CONFIG.exposures.openstreetmap.dict_geofabrik`.

This has the advantages that the code is shorter and easier to digest – and it allows changes in the dictionary in the local config file without changing the code of the module. However, I am not 100% sure whether it should go to config, because it is probably nothing that users would customize very frequently(?)

On the technical side, I am not sure how to write a `dict` in the config file so that it is subscriptable, so that you can call `CONFIG.exposures.openstreetmap.dict_geofabrik[iso3]`

(Here could be a solution: https://stackoverflow.com/questions/38165792/config-file-with-a-dictionary-using-python )

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+    * road, 
+    * rail,  
+    * telecommunications, 
+    * water supply, 
+    * wastewater.
+
+Note: If modified, make sure that key exists in osm.config file, under the
+respective geometry/-ies. 
+"""
+DICT_CIS_OSM = {
```
sameberenz, line 272: Another dictionary that could possible be saved somewhere outside the module code. But maybe ask Emanuel where best to place such dictionaries

sameberenz (commented):
In general, the code makes a lean and readable first impression. Only the long dictionaries in the beginning could be taken somewhere else (either CONFIG or ???).
At this point, I haven't yet looked into functionality and tutorial.

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
chahank: Just as a remark: default values that are never to be changed can be saved as constants in 
https://github.com/CLIMADA-project/climada_python/blob/main/climada/util/constants.py

sameberenz: Thanks Chahan, this could indeed be the right place for these dictionaries.

Evelyn-M:
Thanks for the first remarks!

As a general addition: I think the "second" part, osm_exposuremaker, will end up not being needed. With the lines & polygons functionalities (hopefully) progressing soon in the climada_python repo, there is essentially no reason for osm gdfs to need a separate method for being converted into CLIMADA exposures.

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+import numpy as np
+import itertools
+import time
+import overpy
+
+from climada import CONFIG
+
+LOGGER = logging.getLogger(__name__)
+gdal.SetConfigOption("OSM_CONFIG_FILE",
+                     str(Path(__file__).resolve().parent.joinpath('osmconf.ini')))
```
ChrisFairless, line 38: i'd suggest specifying this path in a config file rather than constructing it within the code. it's less likely to go unexpectedly wrong on weird operating systems and lets people write their own osmconf.ini

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+            'osm_query' : """highway='motorway' or highway='motorway_link' or
+                             highway='trunk' or highway='trunk_link' or
+                             highway='primary' or highway='primary_link' or
+                             highway='secondary' or highway='secondary_link' or
+                             highway='tertiary' or highway='tertiary_link' or
+                             highway='residential' or highway='road' or
+                             highway='service' or highway='unclassified' or
+                             highway='traffic_signals' or
+                             (public_transport='*' and bus='yes') or
+                             man_made='bridge' or man_made='tunnel'"""},
```
ChrisFairless, line 325: will this include other, non-road bridges and tunnels?

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+    * electric power, 
+    * food supply, 
+    * healthcare facilities, 
+    * natural gas infrastructure,
+    * oil infrastructure, 
+    * road, 
+    * rail,  
+    * telecommunications, 
+    * water supply, 
+    * wastewater.
```
ChrisFairless, line 267: are there other things you want to implement? would be good to have a GitHub issue listing what you think would be useful. i think ports would be useful.

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+        
+        Returns
+        -------
+        str : Geofabrik ownload-api for the requested country.
+        
+        See also
+        --------
+        DICT_GEOFABRIK for exceptions / special regions.
+        """
+
```
ChrisFairless, line 410:
```suggestion
         if iso3 not in DICT_GEOFABRIK.keys():
           LOGGER.error('The provided iso3 is not a recognised Geofabrik code')
```

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+            or osm-Protocolbuffer Binary Format (pbf), which is smaller in 
+            size, but has a more complicated query syntax to load (functions
+            are provided in the OSMFileQuery class).
+        save_path : str or pathlib.Path
+            Folder in which to save the file
+            
+        Returns
+        -------
+        None
+            File is downloaded and stored unter save_path + filename
```
ChrisFairless, line 441:
```suggestion
            File is downloaded and stored unter save_path + the Geofabrik filename
```
Also this is very picky, but `save_dir` wouldbe a better variable name than `save_path`

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+        if not Path(save_path).is_file():
+            LOGGER.info(f'Downloading file as {save_path}')
+            urllib.request.urlretrieve(self.planet_url, save_path)
+        else:
+            LOGGER.info(f'file already exists as {save_path}')
+
+    def _osmosis_extract(self, shape, path_planet, path_extract):
+        """
+        Runs the command line tool osmosis to cut out all map info within 
+        shape, from the osm planet file, unless file already exists.
```
ChrisFairless, line 475: Is there a reason why you the default beviour isn't to overwrite an existing file? I would imagine that it's more convenient to be able to overwrite by default (especially if you're extracting to temporary files or files with arbitrary names) than to save people time who accidentally extract the same thing twice.

Maybe we need an `overwrite=False` parameter (or `overwrite=True`, which feels safer to me)

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+        Parameters
+        ----------
+        shape : list or str
+            bounding box [xmin, ymin, xmax, ymax] or file path to a .poly file
+        path_extract : str or pathlib.Path
+            file path (incl. name & ending) under which extract will be stored
+        path_planet : str or pathlib.Path
+            file path to planet-latest.osm.pbf. Will download & store it as 
+            indicated, if doesn`t yet exist.
+            Default is DATA_DIR/planet-latest.osm.pbf
```
ChrisFairless, line 530: Is there any reason I couldn't give this the path to a single-country PBF file that I know contains my bounding box?

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+
+
+        Note
```
ChrisFairless, line 533: Could you add a note here on how to install the osmosis command line tool? (And possibly check for its existence and throw an error if it's not found?)

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+        """
+
+        if not Path(path_extract).is_file():
+            
+            LOGGER.info('File doesn`t yet exist. Assembling osmosis command.')
+            if (isinstance(shape, list) or isinstance(shape, tuple)):
+                cmd = ['osmosis', '--read-pbf', 'file='+str(path_planet), 
+                       '--bounding-box', f'top={shape[3]}', f'left={shape[0]}',
+                       f'bottom={shape[1]}', f'right={shape[2]}', 
+                       '--write-pbf', 'file='+str(path_extract)]
```
ChrisFairless, line 499: Will this run on Windows? 😬 
Could be worth catching errors and suggesting common reasons it fails (you're on Windows, osmosis not installed, etc)

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+        file_format : str
+            Format in which file should be downloaded; options are 
+            ESRI Shapefiles (shp), which can easily be loaded into gdfs,
+            or osm-Protocolbuffer Binary Format (pbf), which is smaller in 
+            size, but has a more complicated query syntax to load (functions
+            are provided in the OSMFileQuery class).
+        save_path : str or pathlib.Path
+            Folder in which to save the file
+            
+        Returns
```
ChrisFairless, line 438: It might be worth returning the file path here. The user won't always know the downloaded file's name and this might speed up someone's workflow. Same applies to other methods in this class.

(Or to save time we could leave this as a GitHub Issue.)

climada_petals/entity/exposures/openstreetmap/osmconf.ini
```diff
@@ -0,0 +1,121 @@
...
+# see http://wiki.openstreetmap.org/wiki/Map_Features
+closed_ways_are_polygons=aeroway=aerodrome,amenity,boundary,building,craft,geological,historic,healthcare=hospital,healthcare=clinic,healthcare=doctors,healthcare=dentist,landuse,leisure,military,natural,office,place,shop,sport,tourism,highway=platform,public_transport=platform,reservoir_type,tower_type,power=generator,power=substation,power=transformer,power=terminal,power=plant,power=tower,man_made=wastewater_plant,man_made=wastewater_tank,utility,water,building=school,amenity=school,building=kindergarten,amenity=kindergarten,building=college,amenity=college,building=university,amenity=university,building=college,amenity=college,building=childcare,amenity=childcare,amenity=hospital,amenity=clinic,amenity=doctors,amenity=pharmacy,amenty=nursing_home,amenity=dentist,telecom=central_office,telecom=data_center,telecom=street_cabinet,railway=platform,railway=station,man_made=storage_tank,man_made=oil_refinery,man_made=petroleum_well,shop=supermarket,shop=grocery,shop=greengrocer,shop=general,shop=bakery,pipeline=substation,public_transport
```
ChrisFairless, line 7: Could we have more context? Did you build this list from that URL? Is it complete (as of 2021), or might I need to add to it?

In general, I'd like have a sense of
(1) when will I have to edit this file (e.g. does this just work for critical infrastructure/common features, or is it complete enough only to need updating when OSM standards change?)
(2) if so, what will i need to change to add/update a feature or customise for my purposes?

If multiple users might be editing this and it would need more than a couple comments to explain, we can create an Issue and document it later.

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+        query = self._query_builder(geo_type, constraint_dict)
+        sql_lyr = data.ExecuteSQL(query)
+        features = []
+        if data is not None:
+            LOGGER.info('query is finished, lets start the loop')
+            for feature in tqdm(sql_lyr, desc=f'extract {geo_type}'):
+                try:
+                    fields = []
+                    for key in ['osm_id', *constraint_dict['osm_keys']]:
+                        fields.append(feature.GetField(key))
```
ChrisFairless, line 651: could use a list comprehension?
```suggestion
                    fields = [feature.GetField(key) for key in ['osm_id', *constraint_dict['osm_keys']]]
```

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+                    geom = shapely.wkb.loads(feature.geometry().ExportToWkb())
+                    if geom is None:
+                        continue
```
ChrisFairless, line 654: what are the situations where geom will be none? is this common?
it feels weird to throw a warning on line 658 when you skip a feature but not here

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+        correct keys / values you are looking for.
+        """
+        driver = ogr.GetDriverByName('OSM')
+        data = driver.Open(self.osm_path)
+        query = self._query_builder(geo_type, constraint_dict)
+        sql_lyr = data.ExecuteSQL(query)
+        features = []
+        if data is not None:
+            LOGGER.info('query is finished, lets start the loop')
+            for feature in tqdm(sql_lyr, desc=f'extract {geo_type}'):
```
ChrisFairless, line 647: this is selfish, but i'd like this to run faster. could you vectorise it?
or log it as an issue for someone to solve later

you might be noticing that i'm a fan of saving nice-but-not-necessary changes as issues

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
Evelyn-M: Thanks for your suggestions! 
So, after a long back and forth, I moved both dictionaries into constants.py with commit cd2ee195.
The reasons being:
1) The geofabrik dictionaries just creates the correct urls. They will most likely hardly change. The one case where users could need to interfere is by adding countries that are on geofabrik, but not in the dict. There might be a few that are still missing (impossible to check manually).
2) The osm_cis_dict might need to be changed, if the user wants to query other key-value pairs under the critical infrastructures than those which I defined. The user can always do this with the more generic osm query functions, too, however.
3) The config file does not allow a nested dict structure (at least not easily), that can then be subscripted. (it's primarily a config object, not a dict, when loaded.). Also, I have structures of the type """Main string with 'string-inside-string' """" patterns. That's not allowed within the config style.

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
Evelyn-M: yes, see above.

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+        shape : list or str
+            bounding box [xmin, ymin, xmax, ymax] or file path to a .poly file
+        path_planet : str or pathlib.Path
+            file path to planet.osm.pbf
+        path_extract : str or pathlib.Path
+            file path (incl. name & ending) under which extract will be stored
+            
+        Returns
+        -------
+        None
```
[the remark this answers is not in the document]
Evelyn-M: I made a more refined if/else statement with LOGGING in commit a7d4b2f7
Now, if there's yet another case other that Path.is_file, or not Path.is_file, it logs this occurrence. In the latter and the first case, it returns None, indeed.

climada_petals/entity/exposures/openstreetmap/osmconf.ini
Evelyn-M: Yes, good point! This does need updating depending on the user's queries. It's by no means complete, I just filled it with whatever was needed for my pre-written CIS-extraction wrappers.
I explain the use of this osm.config.ini file now in the tutorial, with commit 377507b6

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
Evelyn-M: good point. changed this now. 8b95b40d

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
Evelyn-M: Well spotted! I updated the conditions to only report those tunnels & bridges that also have a highway tag.

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
Evelyn-M: Good point. I added it in f0fcbd38, together with a warning for Russia, which is not RUS here.

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
Evelyn-M: At the moment not, but I find it a good idea to have user-feedback on what features would be nice to have an automated query. You can open a GitIssue ;)!

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
Evelyn-M: It is logging the path, tough.  LOGGER.info(f'Downloading file as {local_filepath}') includes the complete file name.

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
Evelyn-M: osmosis_extract takes forever (!) to crop the desired area from the planet file. So you really don't want to mess up. I added a parameter `overwrite` that has a default of False in 243d11b4

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
Evelyn-M: I am not sure how to catch command-line tool errors.
It is supposed to run on windows, when installed 😄

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
Evelyn-M: Good point. No, this should work, equally. I wrote a second more generic function, since this one (with the planet file) might be a convenient wrapper. 26f3026b

Evelyn-M: The only disadvantage is now, that one has to be sure that `shape` is contained within the raw osm.pbf file that is given.

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
Evelyn-M: also added it to the tutorial and several other parts now: 3906d0c0

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
Evelyn-M: You can make it an issue 😄   I actually find it pretty fast (given the other methods of API extraction). But yeah, absolutely no clue how to set up sql queries, and even less so on how to vectorize it. This was really complicated already.

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
Evelyn-M: yeah there are actually instances where there is no geometry reported. I guess for some use cases it could still be useful that a feature is "there" somewhere.. not sure^^

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
Evelyn-M: 2cc20dd3

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+                    print(' WARNING: !!!!\n {}\n try again in {} seconds'.format(exc, waiting_period))
+                else:
+                    raise Exception("The Overpass API is consistently unavailable")
+            time.sleep(waiting_period)
+            waiting_period *= 2
+
+    def _overpass_query_string(self):
+        return f'[out:json][timeout:180];(nwr{self.condition}{self.area};(._;>;););out;'
+
+    def _assemble_from_relations(self, result):
```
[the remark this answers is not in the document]
Evelyn-M: 519e98ef

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
```diff
@@ -0,0 +1,945 @@
...
+
+        # list of lists into list:
+        nodes_taken = list(itertools.chain.from_iterable(nodes_taken))
+        ways_taken = list(itertools.chain.from_iterable(ways_taken))
+
+        return nodes_taken, ways_taken, gdf_rels
+
+    def _assemble_from_ways(self, result, ways_avail, closed_lines_are_polys):
+
+        """assemble gdfs from ways, """
```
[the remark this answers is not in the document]
Evelyn-M: 519e98ef

Evelyn-M:
Thanks for all the comments. I answered them, and one or two should go into GitHubIssues, I believe.
Then the last things remaining:
- A few "special-use" functions need the command-line tool osmosis. For the user who really wants to have those functionalities, it is well documented (both in the functions and in the tutorial), that this needs to be installed. I think this is good enough. As a consequence, I dont see how to test those few funtions though.
- The tutorial's second part (making actual exposures), runs only on climada_python lines_polys_exp branch, since the PR is still pending there. Is this a problem? It's noted in the text there, but i could also comment out the lines, if jenkins checks for notebook execution.
- Some of the functions would need to download (large) files to be tested. Skipping those???

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
ChrisFairless: I feel like that's the user's responsibility. This is also good since you can now cut out a square that includes multiple countries by calling the function multiple times.

ChrisFairless: Or will it behave weird if I do that?

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
ChrisFairless: nice

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
ChrisFairless: i also don't know. maybe @emanuel-schmid can tell us if we should worry.
if it's important we can make an issue

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
ChrisFairless: i'm ok with that. and if for some reason i need the filepath in a script i can get the name from `DICT_GEOFABRIK`

climada_petals/entity/exposures/openstreetmap/osmconf.ini
ChrisFairless: great!

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
ChrisFairless: i wasn't thinking anything fancy - probably converting the `for` loop into a list comprehension by putting the contents into a function. if i need to (try to) speed it up i'll do it :)

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
ChrisFairless: i have another question about this: in the tutorial the geodataframe for zurich buildings includes a row with the geometry GEOMETRYCOLLECTION EMPTY
would this count as `geom is None`? if so we should be dropping the nones in all methods or none

ChrisFairless:
> Thanks for all the comments. I answered them, and one or two should go into GitHubIssues, I believe. Then the last things remaining:
> 
>     * A few "special-use" functions need the command-line tool osmosis. For the user who really wants to have those functionalities, it is well documented (both in the functions and in the tutorial), that this needs to be installed. I think this is good enough. As a consequence, I dont see how to test those few funtions though.
> 
>     * The tutorial's second part (making actual exposures), runs only on climada_python lines_polys_exp branch, since the PR is still pending there. Is this a problem? It's noted in the text there, but i could also comment out the lines, if jenkins checks for notebook execution.
> 
>     * Some of the functions would need to download (large) files to be tested. Skipping those???

this all sounds great!
to test the functions with large downloads ... our options are
- don't
- set up an integration test that downloads a tiny, data-sparse country, and use that when possible (i don't think we can test anything that depends on a  whole-planet download)

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
ChrisFairless: this is good

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
Evelyn-M: right. For the cases where it has a GEOMETRYCOLLECTION EMPTY, one can check with geometry.empty (not `None`, however). I'm doing this in the lines_polys exposuremaker functions (checking, and skipping those for now)

Evelyn-M:
> set up an integration test that downloads a tiny, data-sparse country, and use that when possible (i don't think we can test anything that depends on a whole-planet download)

This sounds reasonable.

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
ChrisFairless: I was thinking, if I already have CH, DE and FR downloaded and want to study a box centred on Basel, can I extract that data with three function calls instead of downloading the planet file?

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
ChrisFairless: if the exposuremaker functions handle this, maybe we just keep the empty geometries here. could be the easiest solution.

climada_petals/entity/exposures/openstreetmap/osm_dataloader.py
Evelyn-M: Right!

Evelyn-M:
So, last update. I added tests, except for those instances that are very hard to test as described above. There is still a test-stub for those in the test-file, such that it could be addressed at a later point.

The tests pass when run from the command line out of the User/climada_petals directory, except for one:
I didn't figure out how to correctly check for raising a KeyError. It does raise the key error as intended, but then I think the check gets confused with the fact that in the output, there's a KeyError+ error message. Help would be appreciated!

Other than that, I'd be happy to include minor feedbacks and merge by Monday, such that it can get into the release.

Evelyn-M:
All written tests are running now.
Collection of remaining remarks is posted in issues.

I'll merge in a bit, if I receive no further comments ;).

Thanks a lot for your review @sameberenz @ChrisFairless !

Evelyn-M:
Upon comments from @chahank I comment out the second part of the tutorial, which demonstrates the end-to-end exposure build with functions from climada_python/feature/lines_polygons_exp branch

PR #6, opened 2021-10-07
bug_fix in crop_production.py

aleeciu:
There seemed to be two issues with crop_production:

1. the assignment of country_codes based on lat and lon crashes unless gridded=True is passed. This is actually an issue of `u_coord.get_country_code()` and I will open it shortly.
2. FAO codes were loaded as alpha3 but then compared to numeric codes, so I changed that and we now load FAO codes as numeric.

3. (Minor: in the tutorial we needed to pass the crop kwarg when loading exposure in order for it to work)

Problem 1 also led `test_crop_production.py ` to crash, and wonder why Jenkins did not capture that.


emanuel-schmid:
Thanks for the fix and the notification. I checked and `test_crop_production` actually does not crash in general. For example, It didn't fail on Jenkins and didn't fail on my Windows computer.

aleeciu:
@emanuel-schmid thanks for checking. You mean it does crash after the fix? or it did not even crash with the previous version?

emanuel-schmid:
@aleeciu the latter, it never crashed

aleeciu:
> @aleeciu the latter, it never crashed

@emanuel-schmid this is really weird. I am using a fresh climada_env on a mac and that test crashes. This however has to do with core climada's util and will open a specific issue about that.

PR #8, opened 2021-10-18
Deprecate set and read constructors in exposures 

zeliest:
Only read_hdf5() is used in Petals (set_from_raster and read_mat are not used).


tovogt:
Looks good, thanks!

PR #16, opened 2021-11-16
Feature/ec codes tc forecast

manniepmkam:
Follow the closed pull request [#249](https://github.com/CLIMADA-project/climada_python/pull/249). The bufr files from ECMWF that contains the TC tracks forecast recently has a file format change (see [here](https://confluence.ecmwf.int/display/FCST/Update+to+Tropical+Cyclone+tracks)). The old version of the TCForecast cannot support the new file format change. Therefore, I propose to use package `ecCodes` maintained by ECMWF themselves to extract the TC track forecast data. `ecCodes` has already been installed in CLIMADA since it is a dependent package of `cfgrib`, no extra installation is needed here.

The `tc_tracks_forecast.py` using `ecCodes` has been rewritten following [this example](https://confluence.ecmwf.int/display/ECC/bufr_read_tropical_cyclone). The unit test `climada_petals/hazard/test/test_tc_tracks_forecast.py` has been updated for the new changes. 

Also a new inegrated test `climada_petals/test/test_tc_forecast_intgr.py` has been added to test the realtime data extraction from ECMWF and also the computation of TC windfield using the forecast tracks.


ThomasRoosli:
Thanks Mannie. That pull request looks great. I removed some trailing white spaces, changed a detail in the integration test to make it work on windows as well and expanded some comments and docstrings. For me the pull request can be merged like this.

ThomasRoosli (approved):
Approved as written in comment before.

climada_petals/hazard/tc_tracks_forecast.py
```diff
@@ -117,6 +121,12 @@ def fetch_ecmwf(self, path=None, files=None):
             except AttributeError:
                 pass
 
+            if os.name == 'nt':
+                try:
+                    file = file.file # if in windows try accessing the underlying tempfile directly incase variable file is tempfile._TemporaryFileWrapper
+                except AttributeError:
+                    pass
```
ChrisFairless, line 128: I think we can avoid a try ... except here:
```suggestion
            if os.name == 'nt' and hasattr(file, 'file'):
                file = file.file # if in windows try accessing the underlying tempfile directly in case variable file is tempfile._TemporaryFileWrapper
```

ChrisFairless: also, what is the code doing here?

ChrisFairless:
Looking through the requirements yml file, I can see we need both `eccodes` and `python-eccodes`. Is there a reason for both? I looked them up on conda forge and they are almost the same

ChrisFairless:
It bugs me that `fetch_bufr_ftp` returns different object types depending on its input parameters - if you're saving the downloaded files it returns a list of filepath strings, and if you're not it returns a list of temporary file objects. (One of Emanuel's fave antipatterns.)

I think that it should _always_ return a list of file objects. This will make it easier to use and easier to test. In the repository it's only used for the file object functionality, so this shouldn't break anything...

I'm happy to implement this if you think it's a good idea.

climada_petals/hazard/tc_tracks_forecast.py
ChrisFairless: actually, this fits into a larger conversation about the module's file handling - we'll chat about it in person!

climada_petals/hazard/tc_tracks_forecast.py
```diff
@@ -188,80 +198,154 @@ def fetch_bufr_ftp(target_dir=None, remote_dir=None):
...
+                            "Track is discarded.", sid)
+            return None
 
-        # setup parsers and querents
-        npparser = pybufrkit.dataquery.NodePathParser()
-        data_query = pybufrkit.dataquery.DataQuerent(npparser).query
+        # ensemble members number
+        ens_no = ec.codes_get_array(bufr, "ensembleMemberNumber")
+
+        # values at timestep 0
```
ChrisFairless, line 242: Is there a reason we don't check the significance value here? I can add it in if useful.
Also ... you have to explain the significance value to me at some point.

climada_petals/hazard/tc_tracks_forecast.py
ChrisFairless: update: we're moving this check inside `read_one_bufr_tc`

climada_petals/hazard/tc_tracks_forecast.py
```diff
@@ -188,80 +198,154 @@ def fetch_bufr_ftp(target_dir=None, remote_dir=None):
...
 
             # subset metadata
-            'wmo_longname': data_query(bufr, '/001027'),
-            'storm_id': data_query(bufr, '/001025'),
-            'ens_type': data_query(bufr, '/001092'),
-            'ens_number': data_query(bufr, '/001091'),
+            'wmo_longname': ec.codes_get(bufr, 'longStormName').strip(),
+            'storm_id': ec.codes_get(bufr, 'stormIdentifier').strip(),
+            'ens_type': ec.codes_get_array(bufr, 'ensembleForecastType'),
+            'ens_number': ec.codes_get_array(bufr, "ensembleMemberNumber"),
```
ChrisFairless, line 334: I'm worrying if this could cause some confusion. Let's say we have track data only for ensemble members 2, 12 and 22. Then we've indexed our results dictionary by as `{0: [...], 1: [...], 2: [...]}` but here in the metadata we're recording the ensemble number as `[2, 12, 22]`. Do we want instead to index results as `{2: [...], 12: [...], 22: [...]}`?

Or is this just not worth worrying about?

manniepmkam: It doesn't matter. The dictionary `{0: [...], 1: [...], 2: [...]}` is only used in this methods to temporarily store the data and later being distribute to the list of xarray dataset (the standard structure of TCTracks. The indices of the dictionary are not returned to the TCTracks class.

ChrisFairless: ahhhh yeah then this didn't matter at all

ChrisFairless:
I've made a few little changes and fixes in 769e53a3aca5d1950e7281ed3947c5d6491f0047

- Add two new parameters to `fetch_ecmwf()`, `target_dir` and `remote_dir` which are passed to `fetch_bufr_ftp`. `target_dir` allows the user to save the raw downloaded forecast data to their disk, and `remote_dir` lets them specify which forecast is downloaded from the FTP.
- Replace a few `try ... catch` statements with `if ... else`
- Modify `fetch_bufr_ftp` to always return a list of open file objects. Previously it might return a list of strings, depending on the input parameters. This makes the method more predictable and the logic a little easier to follow, and means that the files always get closed.
- In `_check_variable` add new parameter `varname`. It's only there to improve the error messages. I wanted this so we could give some advice in the error messages on how unexpected they are (since the code we're adapting seemed fine with filling in missing values for lat and lon, but not other variables, and it's unclear why).
- Add new method `get_value_from_bufr_array` which finds the non-missing value in an extracted BUFR array. This doesn't change any logic, but moves some repeated code into a separate method.
- In `read_one_bufr_tc` create a variable `n_ens = len(ens_no)` which simplifies subsequent code a bit.
- More comments on what the code's doing to help whichever poor person has to edit it next time ECMWF update anything

PR #20, opened 2021-12-06
deprecate set_ls_hist, set_ls_prob, update test & tutorial

Evelyn-M:
As discussed; this changes the two `set_` methods in the Landslide module, updates the tests & tutorial (incl. other now invalid set_ references)


climada_petals/hazard/landslide.py
```diff
@@ -126,11 +127,12 @@ def set_ls_hist(self, bbox, input_gdf, res=0.0083333):
 
         Returns
         -------
-        self : Landslide
+        Landslide : Landslide
             instance filled with historic LS hazard
             set for either point hazards or polygons with specified
             surrounding extent.
         """
+        haz = Landslide()
```
sameberenz, line 135: Why do you initiate the object with `Landslide()` instead of `cls()`?
```suggestion
        haz = cls()
```

climada_petals/hazard/landslide.py
```diff
@@ -220,37 +233,49 @@ def set_ls_prob(self, bbox, path_sourcefile, corr_fact=10e6, n_years=500,
...
+        Landslide : climada.hazard.Landslide instance
             probabilistic LS hazard
 
         See also
         --------
         sample_events_from_probs()
         """
-
+        
+        haz = Landslide()
```
sameberenz, line 244: Why do you initiate the object with `Landslide()` instead of `cls()`?
```suggestion
        haz = cls()
```

sameberenz:
I have only two comments (see above) – besides that it looks good (I assume you checked that the new and old tests are passing)

climada_petals/hazard/landslide.py
Evelyn-M: I'm not sure, but it works equally. I'll change it though!

PR #24, opened 2021-12-06
deprecate set_from_nc

Evelyn-M:
set_method deprecation, as discussed.

Tests & tutorials are running.

1 minor modification (had nothing to do with set_ deprecation): GDP2Asset_converter_2.5arcmin.nc contains the column `if_RF`. Tutorial & gdp_asset.py modules were looking for the name `impf_RF`, which led to keyerror. Converted this back.


climada_petals/entity/exposures/gdp_asset.py
```diff
@@ -240,7 +240,7 @@ def _fast_impf_mapping(countryID, natID_info):
         float,float
         """
     nat = natID_info['ID']
-    impf_RF = natID_info['impf_RF']
+    impf_RF = natID_info['if_RF']
```
zeliest, line 243: Could you try to remove the NatRegIDs.csv file from your climada/data folder and run the code again? This should download the file again, I think that you may have an older version where if was not updated to impf yet. It seems like the column is named 'impf_RF' in the repo version https://github.com/CLIMADA-project/climada_python/blob/main/data/system/NatRegIDs.csv .

zeliest:
Except for the small impf comment, everything looks good :)

climada_petals/entity/exposures/gdp_asset.py
Evelyn-M: Right, converted this back in 6aa44ca

PR #27, opened 2021-12-22
remove ragged nested sequence in np.array

Evelyn-M:
Just a very minor fix to get rid of the following deprecation warning when calling `OSMApiQuery().get_data_overpass`

```
VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
  data=np.array([data_id,data_geom,data_tags]).T,
```


emanuel-schmid:
👍 much nicer that way!

PR #28, opened 2022-01-05
Fix tc_surge_bathtub tests for climada_python#339

tovogt:
This fixes the tests for tc_surge_bathtub according to the changes to the function `read_raster_bounds` in the core repository (https://github.com/CLIMADA-project/climada_python/pull/339).


tovogt:
Since tests are currently failing due to this, we should probably merge this before the next release. Do you agree, @chahank ?

chahank:
I have not really followed the development on this branch, but this fix is good to me. I agree with the merging, thanks!

PR #38, opened 2022-04-08
Develop new Warn class

blassr:
I developed a Warn class which enables the clustering of information (including tests and tutorial). Therefore, I implemented an algorithm based on filtering. The Warn class has a data class as attribute where most information is saved. Warning attribute should not be changed in the current implementation (remove_small_regions method is static).

The algorithm itself is implemented as methods, as this outputs the warning. The classmethod allows to return a warning to the user when instantiating an object. Similar helper functions will be implemented for generating warnings of CLIMADA data (e.g., hazards, see tutorial). Tests with actual hazard data (as in tutorial, MCH example) will be implemented. 

Happy to receive feedback and suggestions for improvements on design of classes and methods.


climada_petals/engine/warn.py
```diff
@@ -0,0 +1,191 @@
...
+            Threshold defining too small regions (number of grid points).
+
+        Returns
+        ----------
+        warning : np.array
+            Warning without too small regions.
+        """
+
+        def increase_levels(warn, size):
+            # increase levels of too small regions to max level of this warning
```
chahank, line 167: This method should be a proper function with docstring.

climada_petals/engine/warn.py
```diff
@@ -0,0 +1,191 @@
...
+            # correct the max_lvl regions generated before down,
+            # until the new regions with it are large enough
+            for i in range(np.max(warn, axis=(0, 1)), np.min(warn, axis=(0, 1)), -1):
+                level = copy.deepcopy(warn)
+                level[warn != i] = 0
+                labels = skimage.measure.label(warn)
+                for j in range(len(np.unique(labels)) + 1):
+                    cnt = np.count_nonzero(labels == j)
+                    if cnt <= size:
+                        warn[labels == j] = i - 1
```
chahank, line 185: This method return nothing... modifying input variables in-place is not a very good idea.

climada_petals/engine/warn.py
```diff
@@ -0,0 +1,191 @@
...
+        def increase_levels(warn, size):
+            # increase levels of too small regions to max level of this warning
+            labels = skimage.measure.label(warn)
+            for i in range(np.max(labels)):
+                cnt = np.count_nonzero(labels == i)
+                if cnt <= size:
+                    warn[labels == i] = np.max(warn, axis=(0, 1))
+            return warn
+
+        def set_new_lvl(warn, size):
```
chahank, line 175: This method is unclear from its name. Make a proper function and add a docstring.

climada_petals/engine/warn.py
```diff
@@ -0,0 +1,191 @@
...
+import numpy as np
+import skimage
+import copy
+from dataclasses import dataclass
+
+LOGGER = logging.getLogger(__name__)
+
+
+@dataclass
+class FilterData:
```
chahank, line 12: Is this class really needed? It looks to me that you never pass an object of this class to a method... Why would you need it then? 

@emanuel-schmid : what do you think, is this dataclass useful?

chahank (changes requested):
This is a great new functionality! I think the next steps are:
- improve the current code and in particular remove some of the anti-patterns
- improve the docstrings and testing of methods 
later on
- add plotting methods
- add saving/loading methods
- add convenience methods for hazard/impact objects

I have not yet checked in details the core of this class, the filtering algorithm as it is currently too hidden inside of the code.

climada_petals/engine/warn.py
```diff
@@ -0,0 +1,191 @@
...
+
+        Returns
+        ----------
+        warn : Warn
+            Generated Warn object including warning
+        """
+        filter_data = FilterData(thresholds, expand, operations, sizes)
+        warn = cls(filter_data, data)
+        data_thrs = warn.threshold_data(data)
+        warn.filter_algorithm(data_thrs)
```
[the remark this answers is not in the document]
blassr:
```suggestion
        warn = cls(filter_data, data)
        data_thrs = warn.threshold_data(data)
        warning = warn.filter_algorithm(data_thrs)
        warn.warning = warning  # does this solve the problem? Or do you mean that the attribute warning should not be changed at all?
```

climada_petals/engine/warn.py
```diff
@@ -0,0 +1,191 @@
...
+        m_thrs : np.array
+            Array of thresholded data
+        """
+        if np.max(data) > np.max(self.filter_data.thresholds) or np.min(data) < np.min(self.filter_data.thresholds):
+            LOGGER.warning('Values of data array are smaller/larger than defined thresholds. '
+                           'Please redefine thresholds.')
+        m_thrs = np.zeros_like(data)
+        for i in range(1, len(self.filter_data.thresholds)):
+            m_thrs[data > self.filter_data.thresholds[i]] = i
+        return m_thrs.astype(int)
```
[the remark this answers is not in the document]
blassr: The returned matrix are the indices of the threshold array, therefore integers. The thresholds itself (thinking about to rename them to bins) do not need to be integers. But with digitize this works without explicit conversion.

climada_petals/engine/warn.py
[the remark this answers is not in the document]
blassr: I think I understand now what you mean and how this should be solved. This suggestion does not solve the problem. Sorry about that.

climada_petals/engine/warn.py
```diff
@@ -0,0 +1,191 @@
...
+        Parameters
+        ----------
+        data : np.array
+            2d np.array containing data to generate warning of.
+        thresholds : np.array
+            Thresholds where data can be binned.
+        expand : Bool
+            Information whether regions of higher level should be expanded with lower levels in filtering algorithm.
+        operations : np.array
+            Type of operations to be applied in filtering algorithm.
```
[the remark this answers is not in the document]
chahank: Why use a numpy array then? Just take a list...

climada_petals/engine/warn.py
```diff
@@ -0,0 +1,238 @@
...
+    """
+    # Explanation of defaults (also maybe transfer to config, talk to Emanuel)
+    OPERATIONS = ['DILATION', 'EROSION', 'DILATION', 'MEDIANFILTERING']
+    SIZES = [2, 3, 7, 15]
+    GRADUAL_DECREASE = True
+    CHANGE_SMALL_REGIONS = True
+    SIZE_TOO_SMALL = 150
+
+    def __init__(self, warn_levels, operations=OPERATIONS, sizes=SIZES,
+                 gradual_decr=GRADUAL_DECREASE, change_sm=CHANGE_SMALL_REGIONS, size_sm=SIZE_TOO_SMALL):
```
chahank, line 43: The init does not need defaults here.

climada_petals/engine/warn.py
```diff
@@ -0,0 +1,238 @@
...
+        Coordinates of warning map.
+    warn_levels : list
+        Warn levels that define the bins in which the input_map will be classified in.
+    """
+
+    # operations that can be applied in filtering algorithm
+    filtering_operations = {
+        "DILATION": lambda bin_m, size: skimage.morphology.dilation(bin_m, skimage.morphology.disk(size)),
+        'EROSION': lambda bin_m, size: skimage.morphology.erosion(bin_m, skimage.morphology.disk(size)),
+        'MEDIANFILTERING': lambda bin_m, size: skimage.filters.median(bin_m, np.ones((size, size)))
```
chahank, line 80: Now that I think of it, maybe instead of using `lambda` methods you could define proper for each operation functions.

blassr:
I worked through Chahan's feedback. Do you want to go through it again, Chahan? Or can Emanuel directly have a look at it?

What needs to be done:

- add saving/loading methods
- add convenience methods for hazard/impact objects

climada_petals/engine/warn.py
emanuel-schmid: The way I understand it, the `FilterData` is just a compilation of Parameters that are used throughout the `Warn.from_map` "constructor".
As such, I think the class has its justification. Storing a set of parameters for later use can be convenient.

However
- the class name may be more to the point, like `WarnParameters` or something
- as it most likely won't be used outside of this particular context one could think about moving it inside the `Warn` class
- though by nature it is most certainly one, it's not implemented as a dataclass. Therefore I suggest to remove that tag or, preferably, implement it as a dataclass (which, among other things, involves importing `List` from `typing`, see https://stackoverflow.com/questions/57690749/how-to-declare-an-array-or-a-list-in-a-python-dataclass)

emanuel-schmid:
👍 I very much like the size of the methods and functions! Nice decomposition. 😃

emanuel-schmid:
And some comments on the tutorial:

-	A bit of text could be useful. Purpose? Means? Limitations?
-	Perhaps one could rename `kwargs` to `plotting_parameters` so it's clear what it is meant for?
-	MCH – what does it stand for?
-	CLIMADA Impact Example -Haiti vs. Impact Example Haiti – what’s the difference between the two and how are they related?
-	One could add semi-colons after plot_warning() at the end of the cell, so it does not list the plot object in the Output.
-	Discrete Warn Levels (categories) while axis labels suggests continuous scaling.
-	`geo_scatter_categorical` plots have unreadable axis labels. Is this on purpose to demonstrate the advantage of using `Warn`?
-	For consistency, why not `**kwargs` instead of `cmap=’Wistia’` in `geo_scatter_categorical` arguments?

emanuel-schmid:
.. and some missing tests according to Jenkins
- wind_mch_default
- plot_warning
- generate_warn_map when there is only one category (which is possibly futile, s.a.)
- Warn.__init__ when sizes and operations have different length and when operations are not allowed

climada_petals/engine/warn.py
blassr: Thanks for that feedback. I made these changes and I also set the defaults now differently (as it is done in dataclasses, right?). Is this better now?

emanuel-schmid (changes requested):

climada_petals/engine/warn.py
```diff
@@ -0,0 +1,367 @@
...
+
+
+from enum import Enum
+
+class Operation(Enum):
+    dilation = dilation
+    erosion = erosion
+    median_filtering = median_filtering
+    
+    def __call__(*args, **kwargs):  # this is not strictly necessary but with this one can do `Operation['dilation'](map, size)`
```
blassr, line 95: Calling like this does not work, or I did something wrong?! 'Operation.dilation(map, size)' does work but we need it to be adjustable. Did I misunderstand something?

climada_petals/engine/warn.py
```diff
@@ -0,0 +1,355 @@
...
+        ----------
+        binned_map : np.ndarray
+            Map of binned values in levels, same shape as input map.
+        """
+        if np.min(input_map) < np.min(levels):
+            LOGGER.warning('Values of input map are smaller than defined levels. '
+                           'Please set the levels lower or check input map.')
+        if np.max(input_map) > np.max(levels):
+            LOGGER.warning('Values of input map are larger than defined levels. '
+                           'Please set the levels higher or check input map.')
```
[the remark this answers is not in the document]
blassr: Yes, it still works. A new higher warn level is generated.

climada_petals/engine/warn.py
```diff
@@ -0,0 +1,355 @@
...
+
+    def __init__(self, warning, coord, filter_data):
+        """Initialize Warn.
+
+        Parameters
+        ----------
+        warning : np.ndarray
+            Warn level for every coordinate of input map.
+        coord : np.ndarray
+            Coordinates of warning map.
```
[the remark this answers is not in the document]
blassr: dito

climada_petals/engine/warn.py
blassr: @emanuel-schmid I would be very happy if you could have a quick look at the newly designed data class. Is this now better and can I improve it further?

chahank:
Can you try to test it on your machine first? I.e., make a fresh CLIMADA environment with the new requirement file and see if the tests run?

blassr:
I just tried that and that doesn't work. Error message:
ModuleNotFoundError: No module named 'skimage'
But I don't know how to solve it correctly.

chahank:
Hmmm, then we should ask @emanuel-schmid

Issue #34, opened 2022-02-18
Faulty frequency definition in LS hazard 


Evelyn-M:
Frequency is defined to have the same shape as intensity in the LS class, instead of being an array of event_id shape. Event_id is also meaningless in the current implementation, as the entire hazard, over whatever yearspan, is just defined as one event.
Needs a re-definition and re-considering of what is an event.

https://github.com/CLIMADA-project/climada_petals/blob/78174b431294dc374cee003e839f1187a9e1443b/climada_petals/hazard/landslide.py#L160

This causes issues during impact calculations with concatenated hazards, for instance.

chahank:
Is there a plan to resolve this issue in the near future?

PR #39, opened 2022-04-12
update intensity & frequency def in probabilistic LS

Evelyn-M:
This solves issue #34, where the Landslide probabilistic hazard had an intensity, frequency and fraction definition which did not go well with the standard hazard definition in climada.

An event is now a year of sampled LS occurrences across the area of interest. Frequency is 1/total sampled years, intensity (binary) and fraction (binary) are csr matrices of shape events (years) * centroids.

Tests and tutorial are amended accordingly


climada_petals/hazard/landslide.py
```diff
@@ -37,43 +37,48 @@
...
-def sample_events_from_probs(prob_matrix, n_years, dist='binom'):
-    """sample an event set for a specified representative time span from
-    a matrix with annual occurrence probabilities
-    Draws events from chosen distribution.
+
+def sample_events(prob_matrix, n_years, dist='binom'):
+    """Repeatedly sample events"""
+    events = []
+    for i in range(n_years):
+        events.append(sample_event_from_probs(prob_matrix, n_years=1, dist=dist))
```
chahank, line 45: Please use list comprehensions for simple for loops.
```suggestion
    events = [
        sample_event_from_probs(prob_matrix, n_years=1, dist=dist)
        for i in range(n_years)
        ]
```

climada_petals/hazard/landslide.py
```diff
@@ -37,43 +37,48 @@
 LOGGER = logging.getLogger(__name__)
 HAZ_TYPE = 'LS'
 
-def sample_events_from_probs(prob_matrix, n_years, dist='binom'):
-    """sample an event set for a specified representative time span from
-    a matrix with annual occurrence probabilities
-    Draws events from chosen distribution.
+
+def sample_events(prob_matrix, n_years, dist='binom'):
+    """Repeatedly sample events"""
```
chahank, line 42: Could you please make a proper docstring? I do not really understand why a sparse matrix is returned, and it is unclear what an event is...

climada_petals/hazard/landslide.py
```diff
@@ -37,43 +37,48 @@
...
+    """sample an event  for a specified representative time span from
+    a matrix with occurrence probabilities. Draws events from chosen 
+    distribution.
 
     Parameters
     ----------
-    prob_matrix : scipy.sparse.csr matrix
-        matrix where each entry has an annual probability [0,1] of occurrence of
+    prob_matrix : np.array()
+        matrix where each entry has a probability [0,1] of occurrence of
```
chahank, line 57: What is the size of the matrix?

climada_petals/hazard/landslide.py
```diff
@@ -37,43 +37,48 @@
...
+
+def sample_events(prob_matrix, n_years, dist='binom'):
+    """Repeatedly sample events"""
+    events = []
+    for i in range(n_years):
+        events.append(sample_event_from_probs(prob_matrix, n_years=1, dist=dist))
+    return sparse.csr_matrix(events)
+    
+    
+def sample_event_from_probs(prob_matrix, n_years=1, dist='binom'):
```
chahank, line 49: This method looks almost identical to some of the methods implemented in `utils.yearsets`. I do not think you need to refactor this here, but we should keep in on the radar for future development.

chahank, line 49:
```suggestion
def sample_event_from_probs(prob_matrix, n_years, dist):
```
If possible, avoid defaults in non-exposed methods. At least I see a potential for conflict as in the method `from_prob` the default is `poisson` which would be at odds here.

climada_petals/hazard/landslide.py
```diff
@@ -229,45 +236,43 @@ def from_prob(cls, bbox, path_sourcefile, corr_fact=10e6, n_years=500,
...
-
-        # set frequ. to no. of occurrences per cell / timespan:
-        haz.frequency = haz.fraction/n_years
-
-        # set intensity to 1 wherever an occurrence:
-        haz.intensity = haz.fraction.copy()
-        haz.intensity[haz.intensity.nonzero()]=1
+        haz.intensity = sample_events(prob_matrix, n_years, dist)
+        haz.fraction = haz.intensity.copy()
+        haz.fraction[haz.intensity.nonzero()]=1
```
chahank, line 261:
```suggestion
        haz.fraction[haz.intensity.nonzero()] = 1
```

climada_petals/hazard/landslide.py
```diff
@@ -229,45 +236,43 @@ def from_prob(cls, bbox, path_sourcefile, corr_fact=10e6, n_years=500,
...
+        haz.frequency = np.ones(n_years)/n_years
```
chahank, line 262:
```suggestion
        haz.frequency = np.ones(n_years) / n_years
```

chahank:
Quick question: I seem to remember that his module had some performance issues? As far as I can see, there is room for improvement of the code if needed, for instance with the handling of sparse matrices.

Evelyn-M:
> > Quick question: I seem to remember that his module had some performance issues? As far as I can see, there is room for improvement of the code if needed, for instance with the handling of sparse matrices.
> 
> ... but this too belongs on the radar for future development and does not need to be addressed within this PR - or does it?

Yep, this is not part of this PR. I was thinking of making another PR to work on the performance issue, as this is also a separate issue on GitHub.

climada_petals/hazard/landslide.py
Evelyn-M: Hope it's clearer now.

climada_petals/hazard/landslide.py
Evelyn-M: I amended the docstring. Let me know if it's understandable.

climada_petals/hazard/landslide.py
```diff
@@ -37,43 +37,88 @@
...
+    """
+    Sample yearly events for a specified time span (n_years) from
+    a matrix representing annual occurrence probabilities per grid cell 
+    (prob_matrix). The matrix will usually be provided throught the
+    Landslide.from_prob() method and refer to grid-cells and their annual 
+    sliding probabilities.
+    Events are drawn from the specified distribution (binomial or 
+    poisson). 
+    Returns n_year events for the whole grid, each event being representative
+    of all slides happening in 1 year throughout the entire grid-area.
```
chahank, line 51: Much improved docstring! Yet, I do not understand what an event is... Is an event a full matrix?

climada_petals/hazard/landslide.py
```diff
@@ -37,43 +37,88 @@
...
+        Landslide.from_prob().
     n_years : int
-        the timespan of the probabilistic simulation in years
-    dist : str
-        distribution to sample from. currently 'binom' (default) and 'poisson'
-
+        the timespan of the probabilistic simulation in years. Each year will
+        result in 1 event.
+    dist : str, 'binom' or 'poisson'
+        distribution to sample from. Default is 'binom'.
```
chahank, line 64: Are the parameters of the binomial and Poisson distribution fixed? Then I think they should be indicated.

climada_petals/hazard/landslide.py
```diff
@@ -37,43 +37,88 @@
...
+    events = [
+        sample_event_from_probs(prob_matrix, n_years=1, dist=dist)
+        for i in range(n_years)
+        ]
+    return sparse.csr_matrix(events)
+    
+    
+def sample_event_from_probs(prob_matrix, n_years, dist):
+    """sample for a specified representative time span  (n_years) from
+    a matrix with occurrence probabilities. Draws from a specified distribution.
```
chahank, line 87: I still do not understand why two methods with the same signature are needed to sample events... But I suppose there is a good reason which is fine with me.

climada_petals/hazard/landslide.py
```diff
@@ -37,43 +37,88 @@
...
+    prob_matrix : np.array()
+        matrix where each entry has a probability [0,1] of occurrence of
+        an event. Shape of array can be any and is determined by hazard area 
+        and resolution of probability-source-file when called through 
+        Landslide.from_prob().
+    n_years : int
+        the timespan of the probabilistic simulation in years. Will be 1 if
+        called from sample_events().
+    dist : str, 'binom' or 'poisson'
+        distribution to sample from.
```
chahank, line 100: Please describe the parameters of the distributions. Binomial or Poisson is not enough.

chahank:
Some small comments on the docstrings. I still do not fully understand the logic of the two added methods, but I see no strong reason to not merge after consideration of the above comments. Thanks @Evelyn-M !

climada_petals/hazard/landslide.py
Evelyn-M: yes, one event refers to all the 0/1s over all grid cells being sampled for that year.

climada_petals/hazard/landslide.py
Evelyn-M: the formula of poisson / binomial distribution sampling is standard knowledge, i'd assume, and can be found explicitly written out in the description to the scipy implementation which is linked below. I added that the only shape params required are n_samples and prob_matrix for clarification (which can be seen in the formula below though, too)

climada_petals/hazard/landslide.py
Evelyn-M: no, they are given by the prob_matrix and n_samples. i added this explicitly to the description now.

climada_petals/hazard/landslide.py
Evelyn-M: i changed this now, indeed from n_years to n_samples in the signature of the 'sample_event_from_probs' methods, to indicate the difference between sampling once n_samples and sampling n_samples times one event

climada_petals/hazard/landslide.py
chahank: I think there might be a misunderstanding. There is no need to describe the functions, but the choice of parameters must be explained. There is no standard for that, and here they are defined under the hood.

Evelyn-M: what parameters do you mean, then? the only ones requried are proability of success (given by prob_matrix) and number of samples to draw from (given by n_samples); all are user-given and explicit

climada_petals/hazard/landslide.py
chahank: I think I am blind. Where are the parameters of binomial and Poisson characterized in the description?

Evelyn-M: Lines #104 / 105

chahank: In lines #104 / 105 I see no mentions of how the Expectation Value is defined.

Evelyn-M: no idea why you need the expectation value in the random variate generator, or how this is relevant information to the user imo. don't think i need to give a crash course on statistics in a docstring, either. i rather think i don't understand what you're hinting at ^^

chahank: Hmmm I think the misunderstanding comes from what is a sample. For instance, for the Poisson distribution and taking the signature of the corresponding scipy method:

`rvs(mu, loc=0, size=1, random_state=None)`

For me:
- size : n_samples
- mu : parameter to be specified (i.e. the expectation value)
- loc : shift of the distribution

If you set `mu = n_samples` as in the code, you will not get back `n_samples` numbers... You will get back one number which gives you the number of events given that the expectation value for events is `mu`. So, that should be reflected in the docstring imho.

Evelyn-M: Oki. I added it accordingly.

PR #41, opened 2022-04-24
Feature/climada entity black marble update

frqqyy:
- Corrected some spelling mistakes
- Updated the link to the documentation of the exposure class
- Changed the code that implements this class (due to the separation of core and petals)
- Corrected the country names with strange symbols


climada_petals/_version.py
```diff
@@ -1 +1 @@
-__version__ = '3.1.1-dev'
+__version__ = '3.1.1'
```
emanuel-schmid, line 1:
```suggestion
__version__ = '3.1.1-dev'
```

setup.py
```diff
@@ -16,7 +16,7 @@
 setup(
     name='climada_petals',
 
-    version='3.1.1-dev',
+    version='3.1.1',
```
emanuel-schmid, line 19:
```suggestion
    version='3.1.1-dev',
```

PR #42, opened 2022-04-26
Fix event_id starting at 0 in RiverFlood

peanutfun:
This fixes a small bug where `event_id` would always start at 0 in `RiverFlood`. According to [the docs](https://climada-python.readthedocs.io/en/stable/climada/climada.hazard.html?highlight=check#climada.hazard.base.Hazard.event_id) of the `Hazard` base class, `event_id` is supposed to be greater zero. The bug prevented, e.g., plotting the hazard intensity of the first event, because `plot_intensity(event=0)` plots the maximum intensity over all events [by convention](https://climada-python.readthedocs.io/en/stable/climada/climada.hazard.html?highlight=check#climada.hazard.base.Hazard.plot_intensity).

There are two minor issues one might like to discuss:
* This problem is not detected by the `check()` method.
* The `check()` method is never executed in the `RiverFlood` unit tests.

### DIFF
--- a/climada_petals/hazard/river_flood.py
+++ b/climada_petals/hazard/river_flood.py
@@ -231,7 +231,7 @@ class RiverFlood(Hazard):
 
         haz.units = 'm'
         haz.tag.file_name = str(dph_path) + ';' + str(frc_path)
-        haz.event_id = np.arange(haz.intensity.shape[0])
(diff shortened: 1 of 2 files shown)


emanuel-schmid:
nice catch! 🎉

PR #43, opened 2022-04-26
Specify data availability for TC emulator tutorial

tovogt:
The description of the data sets that are required to execute the TC emulator tutorial were not clear enough. I clarified the problem about the proprietary license and added my mail address as a reference.


frqqyy:
Thank you very much! It solves the problem perfectly :)

emanuel-schmid:
🙌

PR #46, opened 2022-04-29
fix after change in WIOD data download process

aleeciu:
Fix after the old WIOD16 was archived and thus data download crashed.

### DIFF
--- a/doc/tutorial/climada_engine_SupplyChain.ipynb
+++ b/doc/tutorial/climada_engine_SupplyChain.ipynb
@@ -544,7 +544,7 @@
     "from climada.entity import Exposures\n",
     "\n",
     "countries = ['PHL', 'TWN', 'VNM', 'JPN']\n",
-    "exp_lp = Exposures.concat([client.get_litpop_default(country) for country in countries])\n"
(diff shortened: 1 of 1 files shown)


climada_petals/engine/supplychain.py
```diff
@@ -125,13 +127,21 @@ def read_wiod16(self, year=2014, range_rows=(5,2469),
...
+            LOGGER.info('Downloading folder with WIOD tables')
+
+            downloaded_file_name = u_fh.download_file(WIOD_FILE_LINK,
+                                                      download_dir=WIOD_DIRECTORY)
+            downloaded_file_zip_path = Path(downloaded_file_name + '.zip')
+            Path(downloaded_file_name).rename(downloaded_file_zip_path)
+
+            with zipfile.ZipFile(downloaded_file_zip_path, 'r') as zip_ref:
+                zip_ref.extractall(WIOD_DIRECTORY)
+
```
zeliest, line 142: it seems like the WIOD_DIRECTORY is created when I import the supply chain module. So the check for the directory passes and it is not downloading the file. And further down it fails. So I think that you should rather test if the file exists rather than the directory.

aleeciu:
Many thanks @zeliest and @emanuel-schmid. I am merging this pull request.

PR #49, opened 2022-05-04
Update climada_hazard_Wildfire.ipynb

frqqyy:
- Corrected some spelling mistakes
- Corrected a bug due to the separation of core and petals


emanuel-schmid:
@samluethi : looks like the color scheme has changed, plus setting it in the first cell has no effect. Can you have a look at it?

PR #56, opened 2022-06-21
tc_surge_bathtub: improve performance of sparse matrix operations

tovogt:
This PR greatly improves the performance of the sparse matrix operations in the tc_surge_bathtub hazard module. It reduces the amount of memory required during the computation and for storing the result, and it reduces the computational time considerably. No functionality, and hence no unit tests are affected by these changes.

More details in the comments below.

### DIFF
--- a/climada_petals/hazard/tc_surge_bathtub.py
+++ b/climada_petals/hazard/tc_surge_bathtub.py
@@ -97,11 +97,15 @@ class TCSurgeBathtub(Hazard):
         # Elevation data and coastal/non-coastal indices are used later in the code
         coastal_centroids_h = coastal_centroids_h[elevation_msk]
         coastal_idx = coastal_msk.nonzero()[0]
-        noncoastal_idx = (~coastal_msk).nonzero()[0]
+
+        # to each centroid, assign its position in `coastal_idx`, and assign an out-of-bounds
+        # index to all centroids that are not contained in `coastal_idx`
+        cent_to_coastal_idx = np.full(coastal_msk.shape, coastal_idx.size, dtype=np.int64)
+        cent_to_coastal_idx[coastal_msk] = np.arange(coastal_idx.size)
 
         # Initialize intensity array at coastal centroids
         inten_surge = wind_haz.intensity.copy()
-        inten_surge[:,noncoastal_idx] *= 0
+        inten_surge.data[~coastal_msk[inten_surge.indices]] = 0
         inten_surge.eliminate_zeros()
 
         # Conversion of wind to surge using the linear wind-surge relationship from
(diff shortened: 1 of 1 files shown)


climada_petals/hazard/tc_surge_bathtub.py
```diff
@@ -97,11 +96,15 @@ def from_tc_winds(wind_haz, topo_path, inland_decay_rate=0.2, add_sea_level_rise
...
+
+        # to each centroid, assign its position in `coastal_idx`, and assign an out-of-bounds
+        # index to all centroids that are not contained in `coastal_idx`
+        cent_to_coastal_idx = np.full(coastal_msk.shape, coastal_idx.size, dtype=np.int64)
+        cent_to_coastal_idx[coastal_msk] = np.arange(coastal_idx.size)
 
         # Initialize intensity array at coastal centroids
         inten_surge = wind_haz.intensity.copy()
-        inten_surge[:,noncoastal_idx] *= 0
+        inten_surge.data[~coastal_msk[inten_surge.indices]] = 0
```
tovogt, line 107: The previous code would create a huge dense array in memory. In extreme cases, this array would be so huge that indexing operations on it would fail with a very cryptic error message (`ValueError: could not convert integer scalar`), or the process would be killed due to out-of-memory.

The new code looks complicated, but it's really preventing that any temporary arrays are created that are larger than the amount of non-zero data stored in the sparse matrix.

climada_petals/hazard/tc_surge_bathtub.py
```diff
@@ -119,17 +122,16 @@ def from_tc_winds(wind_haz, topo_path, inland_decay_rate=0.2, add_sea_level_rise
         coastal_centroids_h -= add_sea_level_rise
 
         # Efficient way to subtract from selected columns of sparse csr matrix
-        nz_coastal_cents = inten_surge[:,coastal_idx].nonzero()
-        nz_coastal_cents_inten = (nz_coastal_cents[0], coastal_idx[nz_coastal_cents[1]])
-        inten_surge[nz_coastal_cents_inten] -= coastal_centroids_h[nz_coastal_cents[1]]
+        inten_surge.data -= coastal_centroids_h[cent_to_coastal_idx[inten_surge.indices]]
```
tovogt, line 125: This is a comparably minor improvement, but it's also preventing that any temporary arrays are created that are larger than the amount of non-zero data stored in the sparse matrix.

climada_petals/hazard/tc_surge_bathtub.py
```diff
@@ -119,17 +122,16 @@ def from_tc_winds(wind_haz, topo_path, inland_decay_rate=0.2, add_sea_level_rise
...
         # Discard negative (invalid/unphysical) surge height values
         inten_surge.data = np.fmax(inten_surge.data, 0)
         inten_surge.eliminate_zeros()
 
         # Get fraction of (large) centroid cells on land according to the given (high-res) DEM
-        fract_surge = sp.csr_matrix(_fraction_on_land(centroids, topo_path))
-        fract_surge = sp.csr_matrix(np.ones((inten_surge.shape[0], 1))) * fract_surge
+        # only store the result in locations with non-zero surge height
+        fract_surge = inten_surge.copy()
+        fract_surge.data[:] = _fraction_on_land(centroids, topo_path)[fract_surge.indices]
```
tovogt, line 134: This not only reduces the amount of memory required for computation, but also reduces the effective amount of data stored in the result object. It will only store a non-zero `fraction` value for events and centroids where a non-zero `intensity` value is stored. Since the `fraction` is never used for other centroids, this was a huge waste of storage in the previous version, and would often produce out-of-memory errors.

tovogt (commented):
I explain some of the changes in the code.

tovogt:
@emanuel-schmid Would you please check why some tests are failing that are completely unrelated to the changes introduced in this PR, and that don't fail in the `develop` branch?

emanuel-schmid:
> why some tests are failing that are completely unrelated to the changes introduced in this PR, and that don't fail in the develop branch?

Good news are now they now fail on the `develop` branch too. Bad news dito. 😛

tovogt:
I quickly went through this again and it should be ready to be merged.

emanuel-schmid:
🙌

PR #57, opened 2022-06-21
test_low_flow: increase tolerance when comparing means

emanuel-schmid:
This PR is fixing the small recent discrepancies in the `LowFlow.from_netcdf` means comparison


tovogt:
While it's a bit unsatisfactory that this doesn't address the real cause of this issue, it's definitely safe to restrict the comparison in these specific cases to 6 decimal places. Thanks for the quick fix!

tovogt (approved):

PR #59, opened 2022-09-21
Feature/tc forecast bufr several messages

ThomasRoosli:
_Summary_: I would like to propose a small change that increases the functionality. I leave all the code untouched, except 3 lines plus tests, (it looks much bigger in the code comparison below).
_Justification_: I would like to read in bufr files that contain more than one bufr message per file. This is the case if you get ECMWF bufr files through dissemination or through MARS access.
_Changes_: I want to add `while 1:` in line 249 of tc_tracks_forecast.py and intend the rest of this function `read_one_bufr_tc`. And I add `if bufr is None: break` in line 252 and 253
This follows the python code example provided by ECMWF here: https://confluence.ecmwf.int/display/ECC/bufr_read_tropical_cyclone 

A new testfile and a test is added, to test this new funtionality, it runs on Jenkins


ChrisFairless:
This looks good!

Normally I would get upset about a `while 1:` loop, but since this is taken directly from the ECMWF example it makes sense to keep it.

ChrisFairless:
One unrelated point: when I run the tests on my local machine, the `TestCXML` test fails with
```
FAIL: test_default_xsl (__main__.TestCXML)
"Test with default XSL conversion
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_tc_tracks_forecast.py", line 172, in test_default_xsl
    self.assertEqual(
AssertionError: Timestamp('2022-03-02 12:00:00') != Timestamp('2022-03-02 12:00:00+0000', tz='UTC')

----------------------------------------------------------------------
```
I think this is because the inclusion of timezone information in the value being tested against means that the objects aren't equal.

climada_petals/hazard/test/test_tc_tracks_forecast.py
```diff
@@ -69,10 +70,24 @@ def test_fetch_ecmwf(self):
         self.assertEqual(forecast.data[1].sid, '22S')
         self.assertEqual(forecast.data[1].name, 'HEROLD')
         np.testing.assert_array_equal(forecast.data[0].basin, 'S')
-        self.assertEqual(forecast.data[0].category, 'Tropical Depression')
-        self.assertEqual(forecast.data[0].run_datetime,
-                         np.datetime64('2020-03-19T12:00:00.000000'))
-        self.assertEqual(forecast.data[1].is_ensemble, True)
```
aleeciu, line 75: why were these four lines deleted?

aleeciu: I would expect you need to add a new test but nothing more

aleeciu:
Could you elaborate on what is the `while 1:` meant to do? To my knowledge it's not really good practice in python and maybe we could think of a better way to write it. The example you draw from is in Fortran I guess.

aleeciu:
> One unrelated point: when I run the tests on my local machine, the `TestCXML` test fails with
> 
> ```
> FAIL: test_default_xsl (__main__.TestCXML)
> "Test with default XSL conversion
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "test_tc_tracks_forecast.py", line 172, in test_default_xsl
>     self.assertEqual(
> AssertionError: Timestamp('2022-03-02 12:00:00') != Timestamp('2022-03-02 12:00:00+0000', tz='UTC')
> 
> ----------------------------------------------------------------------
> ```
> 
> I think this is because the inclusion of timezone information in the value being tested against means that the objects aren't equal.

This is something still obscure to me. Many tests crash on the local machine, I believe it has to do with the location tests are ran from. Anyway, as long as Jenkins is happy it should be fine.

climada_petals/hazard/test/test_tc_tracks_forecast.py
ThomasRoosli: You are absolutely right. this is a mistake.

ThomasRoosli:
Regarding `while 1:`
You can click on a "rider" above the Fortran code to switch to the python code example, this is where this code template is from. https://confluence.ecmwf.int/display/ECC/bufr_read_tropical_cyclone
It should try to read from the provided bufr file message by message. It is unclear how many messages there are in the beginning. As soon as there are no other messages it breaks the while loop. I could introduce a variable `check_for_messages=True` before the loop and set it to False when there are no longer any messages. Would you prefer that?

climada_petals/hazard/test/test_tc_tracks_forecast.py
ThomasRoosli: The lines have been reintroduced now.

PR #61, opened 2022-10-27
update __init__ petals Warn

manniepmkam:


emanuel-schmid:
Is this still a valid PR? 
If so I'd need some motivation for changing the order of arguments.
Otherwise I'll close it un-merged. 
Please object if you want to follow up on it, @manniepmkam

emanuel-schmid:
staled out

PR #81, opened 2023-06-08
Integrate the BOARIO model into CLIMADA SupplyChain Module

aleeciu:
This PR integrates the BOARIO model (https://github.com/spjuhel/BoARIO) developed by @spjuhel into the CLIMADA's SupplyChain module.

Missing:
- test
- tutorial
- add boario dependence
- fix https://github.com/CLIMADA-project/climada_petals/issues/113

### PR Author Checklist

#### TODOs:
- [x] Make sure to set BoARIO dependency to ``>=v0.5.9`` for compatibility. 
- [x] Make sure v0.5.9 is made available on conda-forge.

### PR Reviewer Checklist

-----
23/04/24 Remarks from @spjuhel while working on the PR:
 
- Doing multiple run with different event type can lead the ``boario_params["event"]`` dict to hold unwanted keyword arguments for specific events (ie ``boario_params["event"]["recovery_time"]`` is set to a default value if we instantiate an ``EventKapitalRecovery``, which currently causes an error when instantiating an ``EventKapitalRebuild`` afterward. I've corrected this on the boario side, which will be part of its v0.5.9 release.
- Also, on this point, I'm actually a bit unsure if setting default values and logging it in debug is the proper way. I would at least change it to a warning, and I'm more in favour of raising an explicit error that tells the user to set a value.
- ``pymrio`` v0.5.4, just out, introduces Ghosh calculation (which we currently do by ourselves). I think we can just bump the dependency to this version and use their functions.
- The objective of the PR is to finish the coupling with boario, but **I plan to do a rework of the module soon**, to improve its overal structure, test coverage, etc...


spjuhel:
Hi @aleeciu,

This is handled by the type of Event you instantiate. Currently there's 3:
- EventKapitalRecover (Currently in Climada, it was the simplest + most tested one at the time)
- EventKapitalRebuild (Where the direct damages also create a rebuilding demand)
- EventArbitraryProd (Which I implemented recently and corresponds to case 3)

Note that:
- Case 3 has been far less extensively tested and used than case 1 and 2.
- `impact` has to be given as a fraction of production capacity directly in this case.

I think in the long run, Climada's default should probably be EventKapitalRebuild, as it is hard to justify the absence of a rebuilding demand (especially after a Cyclone, for instance). But that requires to define the set of rebuilding sectors and their "market share" of the rebuilding demand, which is hard to generalize over different MRIOT.

climada_petals/engine/supplychain.py
```diff
@@ -568,15 +601,21 @@ def calc_impacts(self,
                     io_approach,
                     exposure=None,
                     impact=None,
-                    impacted_secs=None):
+                    impacted_secs=None,
+                    shock_factor=None,
+                    boario_params=dict(),
```
emanuel-schmid, line 606: suggested way for empty `dicts` as default: ```python
def f(a=None):
    if a is None:
        a = {}
```

**Comment by emanuel-schmid** [comment]
the package `boario` from pypi has a dependency restriction of `pandas<2.0`, (we're at 2.1 atm) which makes me somewhat uneasy. 😟 It doesn't seem to break anything (yet) but I wonder whether there is a performance impact.

**Comment by spjuhel** [comment]
I found a bug in multi-events simulation which I had to fix quickly, so I actually got back into the code earlier than I thought. I used the occasion to loosen the version restriction for `pandas`. My integration tests all passed using `pandas 2.1.3` so this should be fine.

**Comment by emanuel-schmid** [comment]
> I used the occasion to loosen the version restriction for `pandas`.

Awesome. Many thanks!

**Comment by peanutfun** [comment]
In the latest Petals release we had some issues with requirements not being available from conda-forge. We could not create a Conda package that would require packages from both conda-forge and PyPI. @spjuhel @aleeciu May I ask you to add BOARIO to conda-forge? Only then we are actually able to merge this PR and include it in a future release. With tools like `grayskull`, creating the conda-forge is actually quite easy, see https://conda-forge.org/docs/maintainer/adding_pkgs/

**Comment by spjuhel** [comment]
How urgent is this ? (Can it wait 1 month or so for instance ?)

**Comment by davidnbresch** [comment]
I guess it can wait - might be worth checking who brought this upProf. Dr. David N. Bresch Professor for Weather and Climate RisksInstitute for Environmental DecisionsETH ***@***.***@meteoswiss.choffice: +41 44 632 77 87mobile: +41 79 770 15 92skype: davidnbreschzoom: https://ethz.zoom.us/my/dbreschhttps://www.usys.ethz.ch/en/people/profile.david-bresch.html, http://www.wcr.ethz.chOn 21 Feb 2024, at 21:41, Samuel Juhel ***@***.***> wrote:﻿
How urgent is this ? (Can it wait 1 month or so for instance ?)

**Comment by chahank** [comment]
@spjuhel : have you manage to make a conda release for BOARIO?

**Comment by chahank** [comment]
@aleeciu : Can you please provide a rough timeline for finishing this PR?

**Comment by aleeciu** [comment]
> @aleeciu : Can you please provide a rough timeline for finishing this PR?

I guess ownership on this PR now shifted to @spjuhel so I would leave it up to him decide scope and timeline. From where I left, the coupling worked and only tests had to written.

**Comment by spjuhel** [comment]
> @spjuhel : have you manage to make a conda release for BOARIO?

Yes, there's a conda release now!

> Can you please provide a rough timeline for finishing this PR?

I need to look a bit deeper into it, but if it is only a matter of writing tests it can probably be done quickly (by the end of next week seems _a priori_ reasonable to me). Are there specific time constraints? What would be ideal?

**Comment by chahank** [comment]
> > @spjuhel : have you manage to make a conda release for BOARIO?
> 
> Yes, there's a conda release now!

Awesome!

> > Can you please provide a rough timeline for finishing this PR?
> 
> I need to look a bit deeper into it, but if it is only a matter of writing tests it can probably be done quickly (by the end of next week seems _a priori_ reasonable to me). Are there specific time constraints? What would be ideal?

There is no specific time constraints. It would just be good to not let older PR remain open and never be merged. I think here a) tests need to be written b) a tutorial/documentation should be written c) a fix seems open, see the top comment d) a review should be done. So, I fear this might not be a quick quick thing, but also not very long.

**Comment by spjuhel** [comment]
- [ ] Make sure to set BoARIO dependency to ``>=v0.5.9`` for compatibility. 
 
Doing multiple run with different event type can lead the ``boario_params["event"]`` dict to hold unwanted keyword arguments for specific events (ie ``boario_params["event"]["recovery_time"]`` is set to a default value if we instantiate an ``EventKapitalRecovery``, which currently causes an error when instantiating an ``EventKapitalRebuild`` afterward. I've corrected this on the boario side, which will be part of the v0.5.9 release.

Also, on this point, I'm actually a bit unsure if setting default values and logging it in debug is the proper way. I would at least change it to a warning, and I'm more in favour of raising an explicit error that tells the user to set a value.

**Comment by spjuhel** [comment]
@emanuel-schmid or @peanutfun this is almost done, I have some quick technical questions:
I have to:
- [ ] enforce the v0.5.9 dependency for boario (which is available on conda-forge): This should be in env_climada.yml right?
- [ ] solve these conflicts in the requirements: as both pymrio and boario are on conda-forge, can I remove them from setup.py?
- [ ] update the changelog: any do and don't for this?.

File: climada_petals/engine/supplychain.py [hunk 2]
```diff
@@ -99,28 +100,49 @@ def calc_B(Z, x):
...
+def calc_G(B):
+    """Calculate the Ghosh inverse matrix G either from B
+    G = inverse matrix of (I - B) where I is an identity matrix of same shape as B.
+    Note that we define G as the transpose of the Ghosh inverse matrix, so that we can apply the factors of
+    production intensities from the left-hand-side for both Leontief and Ghosh attribution. In this way the
+    multipliers have the same (vector) dimensions and can be added.
+    Parameters
+    ----------
+    B : pandas.DataFrame or numpy.array
+        Symmetric input output table (coefficients)
```
ValentinGebhart, line 126: Why does B have to be symmetric?

climada_petals/engine/supplychain.py
```diff
@@ -99,28 +100,49 @@ def calc_B(Z, x):
...
+    -------
+    pandas.DataFrame or numpy.array
+        Ghosh input output table G
+        The type is determined by the type of B.
+        If DataFrame index/columns as B
+    """
+    I = np.eye(B.shape[0])
+    if isinstance(B, pd.DataFrame):
+        return pd.DataFrame(
+            np.linalg.inv(I - B), index=B.index, columns=B.columns
```
ValentinGebhart, line 137: Shouldn't inverting switch the index labels with column labels? Maybe they are identical here?

climada_petals/engine/supplychain.py
```diff
@@ -453,9 +525,9 @@ def calc_shock_to_sectors(self,
                               shock_factor=None
                               ):
         """Calculate exposure, impact and shock at the sectorial level.
-        This function translate spatially-distrubted exposure and impact
-        information into exposure and impact of MRIOT's country/sectors and
-        for each hazard event.
+        This function translate spatially-disturbed exposure and impact
```
ValentinGebhart, line 528: distributed?

climada_petals/engine/supplychain.py
```diff
@@ -604,8 +711,15 @@ def calc_impacts(self,
 
         self.calc_matrices(io_approach=io_approach)
 
+        # This block is a temporary `fix`, before a deeper rework of the module.
+        # The underlying problem is that if secs_shock was set outside this method,
+        # impact and exposure are ignored as ``calc_shock_to_sectors`` is not run in the next block.
+        # The objective here is to warn user.
+        if (exposure is not None) or (impact is not None) and self.secs_shock is not None:
```
ValentinGebhart, line 718: don't you need a bracket here? ( ... or ...) and ...
"and" has higher precedence than "or"

climada_petals/engine/supplychain.py
```diff
@@ -635,16 +749,144 @@ def calc_impacts(self,
                 axis=1,
             ).T.set_index(self.secs_shock.index)})
 
-        elif io_approach == "eeioa":
-            self.supchain_imp.update({io_approach : (
-                pd.DataFrame(
-                    self.secs_shock.dot(self.inverse[io_approach])
-                    * self.mriot.x.values.flatten()
-                )
-            )})
```
ValentinGebhart, line 752: I don't understand what happens if io_approach is set to "eeioa"? This is not treated in the if statement anymore?

climada_petals/engine/supplychain.py
spjuhel: B (and A), are computed from Z, the transaction matrix (and share the same shape).
Z represent the flow from each (sector, region) to each (sector, region), thus is symmetric.
It doesn't have to be, it is. The word symmetric is descriptive rather than normative here. I could remove it.

Users of the module should know the basics of IO theory, so it doesn't matter much I think.

climada_petals/engine/supplychain.py
spjuhel: Yes they are identical

climada_petals/engine/supplychain.py
spjuhel: Indeed, nice catch!

climada_petals/engine/supplychain.py
spjuhel: Yes, I decided to remove this option.
It was developed by Msc students working with Alessio and reading their thesis I found it very unconvincing. There's absolutely no scientific literature supporting it, and it is a "weird" mashup of Economic Footprint and risk analysis.

I decided that as it was not possible to properly document it and explain it in the tutorial, it should be removed until then.

climada_petals/engine/supplychain.py
spjuhel: Indeed!

climada_petals/engine/supplychain.py
aleeciu: they haven't worked with me :-) .. fully aligned with your assessment @spjuhel ...

climada_petals/engine/supplychain.py
aleeciu: i would leave it, if i am not mistaken it's also aligned with how `pymrio` docstring reports them

doc/tutorial/climada_engine_SupplyChain.ipynb
```diff
@@ -8,9 +8,135 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This tutorial shows how to use the `SupplyChain` class of CLIMADA. This class allows assessing indirect economic impacts via Input-Output (IO) based modeling. \n",
+    "\n",
+    "This tutorial assumes you are familiar with direct impact computation with CLIMADA with the [`Exposures`](https://climada-python.readthedocs.io/en/stable/tutorial/climada_entity_Exposures.html), [`Hazard`](https://climada-python.readthedocs.io/en/stable/tutorial/climada_hazard_Hazard.html) and [`Impact`](https://climada-python.readthedocs.io/en/stable/tutorial/climada_engine_Impact.html) classes. Likewise, this tutorial assumes you are familiar with IO based modeling¹.\n",
+    "\n",
+    "¹) We recommand the following book: Miller, R. E., & Blair, P. D. (2009). Input-Output Analysis: Foundations and Extensions (2nd ed.). Cambridge: Cambridge University Press. "
```
ValentinGebhart, line 26: We recommend ...

spjuhel:
I think I have adressed all remaining problems and remarks and the last remaining thing would be to add `boario==0.5.9` dependency on jenkins (@emanuel-schmid can you do it?), and check that the tests are running.

PR #85, opened 2023-07-05
Implement a new tropical cyclone rain model

tovogt:
Changes proposed in this PR:
- This implements a new model (called **TCR**) as part of the `climada_petals.hazard.tc_rainfield` module.

This PR is still work in progress, but should be finished by the end of this week. I will add a tutorial notebook and some more tests.

### PR Author Checklist

### PR Reviewer Checklist

### DIFF
--- a/climada_petals/hazard/tc_rainfield.py
+++ b/climada_petals/hazard/tc_rainfield.py
@@ -564,11 +564,11 @@ def _compute_rain_sparse(
     rainrates_sparse = None
     if store_rainrates:
         n_reachable_coastal_centr = reachable_coastal_centr_idx.size
-        indices = np.zeros((npositions, n_reachable_coastal_centr, 2), dtype=np.int64)
-        indices[:, :, 0] = 2 * reachable_coastal_centr_idx[None]
-        indices[:, :, 1] = 2 * reachable_coastal_centr_idx[None] + 1
-        indices = indices.ravel()
-        indptr = np.arange(npositions + 1) * n_reachable_coastal_centr * 2
+        indices = np.broadcast_to(
+            reachable_coastal_centr_idx[None],
+            (npositions, n_reachable_coastal_centr),
+        ).ravel()
+        indptr = np.arange(npositions + 1) * n_reachable_coastal_centr
         rainrates_sparse = sparse.csr_matrix((rainrates.ravel(), indices, indptr),
                                               shape=rainrates_shape)
         rainrates_sparse.eliminate_zeros()
--- a/climada_petals/hazard/tc_rainfield.py
+++ b/climada_petals/hazard/tc_rainfield.py
@@ -34,9 +34,9 @@ import xarray as xr
 
 from climada.hazard import Hazard, TCTracks, TropCyclone, Centroids
 from climada.hazard.trop_cyclone import (
-    _close_centroids,
-    _compute_angular_windspeeds,
-    _track_to_si,
+    get_close_centroids,
+    compute_angular_windspeeds,
+    tctrack_to_si,
     GRADIENT_LEVEL_TO_SURFACE_WINDS,
     H_TO_S,
     KM_TO_M,
@@ -90,10 +90,10 @@ M_DRY_AIR = 28.9634
 """Molar mass of dry air (in g/mol)"""
 
 R_GAS = 8.3144621
-"""Molar gas constant (in J / molK)"""
+"""Molar gas constant (in J/molK)"""
 
 R_DRY_AIR = 1000 * R_GAS / M_DRY_AIR
-"""Specific gas constant of dry air (in J / kgK)"""
+"""Specific gas constant of dry air (in J/kgK)"""
 
 RHO_A_OVER_RHO_L = 0.00117
 """Density of water vapor divided by density of liquid water"""
@@ -671,7 +671,7 @@ def compute_rain(
     # convert track variables to SI units
     si_track = _track_to_si_with_q_and_shear(track, **model_kwargs)
 
-    # normalize longitude values (improves performance of `dist_approx` and `_close_centroids`)
+    # normalize longitude values (improves performance of `dist_approx` and `get_close_centroids`)
     u_coord.lon_normalize(centroids[:, 1], center=si_track.attrs["mid_lon"])
 
     # Filter early with a larger threshold, but inaccurate (lat/lon) distances.
@@ -681,7 +681,7 @@ def compute_rain(
     )
 
     # restrict to centroids within rectangular bounding boxes around track positions
-    track_centr_msk = _close_centroids(
+    track_centr_msk = get_close_centroids(
         si_track["lat"].values,
         si_track["lon"].values,
         centroids,
@@ -713,7 +713,7 @@ def compute_rain(
     if model == MODEL_RAIN["R-CLIPER"]:
         rainrates = _rcliper(si_track, d_centr[""], close_centr_msk, **model_kwargs)
     elif model == MODEL_RAIN["TCR"]:
-        rainrates = _tcr(si_track, centroids, d_centr, close_centr_msk, **model_kwargs)
+        rainrates = _tcr(si_track, track_centr, d_centr, close_centr_msk, **model_kwargs)
     else:
         raise NotImplementedError
     [reachable_centr_idx] = track_centr_msk.nonzero()
@@ -792,7 +792,7 @@ def _track_to_si_with_q_and_shear(
         TC track data.
     q_900 : float, optional
         If the track data does not include "t600" values, assume this constant value of saturation
-        specific humidity (in gm/gm) at 900 hPa. Default: 0.01
+        specific humidity (in kg/kg) at 900 hPa. Default: 0.01
     matlab_ref_mode : bool, optional
         Do not apply the fixes to the reference MATLAB implementation. Default: False
     kwargs : dict
@@ -802,7 +802,7 @@ def _track_to_si_with_q_and_shear(
     -------
     xr.Dataset
     """
-    si_track = _track_to_si(track)
+    si_track = tctrack_to_si(track)
 
     if "q900" in track.variables:
         si_track["q900"] = track["q900"].copy()
@@ -857,7 +857,7 @@ def _centr_distances(
     Parameters
     ----------
     si_track : xr.Dataset
-        TC track data in SI units, see `_track_to_si`.
+        TC track data in SI units, see `tctrack_to_si`.
     centroids : ndarray
         Each row is a pair of lat/lon coordinates.
     metric : str, optional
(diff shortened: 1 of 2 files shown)


climada_petals/hazard/tc_rainfield.py
```diff
@@ -66,204 +208,1449 @@ def set_from_tracks(self, *args, **kwargs):
...
-        description : str, optional
-            description of the events
+            Centroids where to model TC. Default: global centroids at 360 arc-seconds resolution.
         pool : pathos.pool, optional
-            Pool that will be used for parallel computation of wind fields. Default: None
+            Pool that will be used for parallel computation of rain fields. Default: None
+        description : str, optional
+            Description of the event set. Default: "".
+        model : str, optional
+            Parametric rain model to use: "R-CLIPER" (faster and requires less inputs, but
```
bguillod, line 255: I would suggest that references for each model are provided here as well. I see that they are provided in the individual method for each model, but for the normal user it might be useful to have them in the documentation of this main method.

tovogt: Thanks! I added more details about the models to the docstring of `TCRain.from_tracks`.

tovogt:
Hey everyone, the new rain module is ready for review!

Please note that running the notebook and the tests will require data sets that are included in the `./data/system/` and `./data/demo/` directories of the repository. If you struggle with error messages about missing data files, you might need to copy those files to the CLIMADA data and config directory on your computer.

climada_petals/hazard/tc_rainfield.py
```diff
@@ -66,204 +215,1562 @@ def set_from_tracks(self, *args, **kwargs):
...
+            track.isel(time=slice(chunk_start, chunk_end)), *args,
+            max_memory_gb=max_memory_gb, **kwargs,
+        )
+        intensity.append(inten)
+        rainrates.append(rainr)
 
-    # Define Coefficients (CLIPER NHC bias adjusted (Tuleya, 2007))
-    a1 = -1.1  # inch per day
-    a2 = -1.6  # inch per day
-    a3 = 64.   # km
```
tovogt, line 244: There was a typo in this line that is fixed in this PR. The value for a3 is 64.5 instead of 64.0, see Table 2 in Tuleya et al. 2007:
image
That's why the intensity values in the unit tests have changed.

climada_petals/hazard/test/test_tc_rainfield.py
```diff
@@ -65,28 +75,48 @@ def test_set_one_pass(self):
...
 
-        self.assertAlmostEqual(tc_haz.intensity[0, 100], 99.7160586771286, 6)
-        self.assertAlmostEqual(tc_haz.intensity[0, 260], 33.2087621869295)
-        self.assertEqual(tc_haz.fraction[0, 100], 1)
-        self.assertEqual(tc_haz.fraction[0, 260], 1)
+        self.assertTrue(isinstance(tc_haz.intensity, sparse.csr_matrix))
+        self.assertEqual(tc_haz.intensity.shape, (1, 296))
+        self.assertEqual(tc_haz.intensity.nonzero()[0].size, 296)
+        self.assertAlmostEqual(tc_haz.intensity[0, 100], 98.61962462510677, 6)
+        self.assertAlmostEqual(tc_haz.intensity[0, 260], 27.702589231065055)
```
tovogt, line 86: Changing the R-CLIPER model was not the purpose of this PR. Still, as mentioned in the comments above we found and fixed a bug in line 244 of the implementation along the way. That's why we also had to change the intensity values in the unit tests for the R-CLIPER model.

climada_petals/hazard/test/test_tc_rainfield.py
```diff
@@ -97,55 +127,64 @@ def test_from_file_pass(self):
...
+        )
 
         rainfall = np.round(rainfall, decimals=9)
 
-        self.assertAlmostEqual(rainfall[0, 0], 66.801702386)
-        self.assertAlmostEqual(rainfall[0, 130], 43.290917792)
-        self.assertAlmostEqual(rainfall[0, 200], 76.315923838)
+        self.assertAlmostEqual(rainfall[0, 0], 65.114948501)
+        self.assertAlmostEqual(rainfall[0, 130], 39.584947656)
+        self.assertAlmostEqual(rainfall[0, 200], 73.792450959)
```
tovogt, line 184: This change in intensity values for the R-CLIPER model is also due to the bug fix described in the comments above.

climada_petals/hazard/tc_rainfield.py
aleeciu: good, tahnks. I think it was me implementing the original R-CLIPER, but I mostly translated the CLIMADA MATLAB code without looking into the refs.

climada_petals/hazard/tc_rainfield.py
```diff
@@ -65,200 +214,1552 @@ def set_from_tracks(self, *args, **kwargs):
...
+        storm center. For "u850"/"v850", use the average over the 200-500 km annulus around the
+        storm center. If "u850"/"v850" is missing, this implementation sets the shear component of
+        the vertical velocity to 0. If "t600" is missing, the saturation specific humidity is set
+        to a universal estimate of 0.01 kg/kg. Both assumptions can have a large effect on the
+        results (see Lu et al. 2018).
+
+        The implementation of the R-CLIPER model currently does not allow modifications, so that
+        `model_kwargs` is ignored with `model="R-CLIPER"`. While the TCR model can be configured in
+        several ways, it is usually safe to go with the default settings. Here is the complete list
+        of `model_kwargs` and their meaning with `model="TCR"` (in alphabetical order):
```
aleeciu, line 268: I'd expect the explanation of what _model_kwargs_ is about in the function's parameters description just below, under _model_kwargs_.

climada_petals/hazard/tc_rainfield.py
```diff
@@ -65,200 +214,1552 @@ def set_from_tracks(self, *args, **kwargs):
...
+        e_precip : float, optional
+            Precipitation efficiency (unitless), the fraction of the vapor flux falling to the
+            surface as rainfall (Lu et al. 2018, eq. (14)). Note that we follow the MATLAB
+            reference implementation and use 0.5 as a default value instead of the 0.9 that was
+            proposed in Lu et al. 2018. Default: 0.5
+        elevation_tif : Path or str, optional
+            Path to a GeoTIFF file containing digital elevation model data (in m). If not
+            specified, a topography at 0.1 degree resolution provided with CLIMADA is used.
+            Default: None
+        matlab_ref_mode : bool, optional
```
aleeciu, line 282: Why do we need this. If you are sure it's a bug, it's better if it's gone, no? I'd rather add in the docstring that this is the original implementation, with a bug fix, and explain what the bug fix is.

aleeciu:
Thanks Thomas, I screened through the code and ran the tutorial. I cannot check it thoroughly but I wanted to provide a user-type of feedback.

Code: It looks great, I just had a couple of documentation-related comments.

Tutorial: I could run itl but I had to specifiy the paths of the required additional files (_c_drag_500_, _topography_land_360_ and _tcrain_examples_) via the _model_kwargs_ arguments as my machine was looking for data under the 'Users/user_name/climada/...' folder instead of the CLIMADA project's folder. 

I am not completely sure, but my understanding is that 'Users/user_name/climada/...' is where these data are supposed to be by default? Anyway, beside that, it ran flawlessly and the text is also very clear with a good balance of explanation and links for further knowledge.

A couple of more points :

- What is the origin of topo data? would be great to know (haven't seen that, but perhaps is there)

- It seems this new model's results are waaay different than R-CLIPER, should just kill R-CLIPER for good or we still want to keep both?

- My understanding is that this code is not applicable to the Knutson-scaled future TCs, right? But what about scaling the generated rainfields with knutson's rain rate scaling factors? would that be sensible ?

tovogt:
Thanks for your feedback!

* Well spotted, I forgot to add the data files to the `REPO_DATA` in `./climada_petals/__init__.py`. This should be fixed now and the files will be automatically copied over to the right locations during import.
* The topography data is taken from the MATLAB reference implementation. The original authors were not able to trace back where this data came from. I tried to replicate the data based on SRTM-upscaling, but in vain. The advantage of this data set is that it is extremely small (~3 MB) so that it can be included in the CLIMADA repository. When upscaling SRTM data to 0.1 degree resolution, the file is much bigger (I don't understand why, actually). There might be a little licensing issue when including data of unknown origin in the repository. But at the same time, it's really a very small, low-resolution topography data set. I don't think that anyone really cares about it...
* I would keep R-CLIPER since it's really very simple code (not hard to maintain) and nice for future reference.
* I haven't worked with the Knutson scaling that is currently implemented in CLIMADA, so I cannot really comment on that.

climada_petals/hazard/tc_rainfield.py
tovogt: I agree that bugs should just be fixed if we are sure that they are bugs. I updated the description of the parameter.

Background: When I originally implemented this, I found quite a few strange "issues" in the MATLAB code. And it took some time to write them down and get feedback from the original authors about these "issues". I didn't want to remove the issues unless the original authors definitely confirm that they are unintended ("bugs"). When I wrote this docstring I still didn't have feedback from the authors. Today, the situation is different: I know that some of the issues really were bugs. Those bugs are now fixed in my implementation and the original authors also decided to fix them in their implementation. But some of the "issues" were actually deliberate decisions by the original authors that I, personally, don't like (for example, using a parameter value that is not documented anywhere in the literature, not even in their own papers), and, at the same time, only make a comparably small difference. For testing purposes, I still think that it's useful to be able to replicate the behavior of the reference implementation. That's why I keep this parameter, but make clear what it actually does.

climada_petals/hazard/tc_rainfield.py
tovogt: Good point, I moved the description accordingly.

aleeciu:
> Thanks for your feedback!
> 
> * Well spotted, I forgot to add the data files to the `REPO_DATA` in `./climada_petals/__init__.py`. This should be fixed now and the files will be automatically copied over to the right locations during import.
> * The topography data is the one from the MATLAB reference implementation. which is just SRTM upscaled to 0.1 degree resolution. The advantage of this data set is that it is extremely small (~3 MB) so that it can be included in the CLIMADA repository. Since SRTM is public domain, there shouldn't be a licensing issue.
> * I would keep R-CLIPER since it's really very simple code (not hard to maintain) and nice for future reference.
> * I haven't worked with the Knutson scaling that is currently implemented in CLIMADA, so I cannot really comment on that.
> 
> EDIT: I was finally able to reproduce the topography file from SRTM using gdalwarp and the "average" interpolation option. This is now also mentioned in several docstrings:
> 
> https://github.com/CLIMADA-project/climada_petals/blob/d649a5adb367582d70cf87bde0b2f8b766dc2104/climada_petals/hazard/tc_rainfield.py#L111-L115

Super ! thanks for addressing these points and especially for figuring out where the elevation data come from. All good for me regarding this PR.

tovogt:
Yes, sounds great! Please go ahead and copy the files to the repository and tell me how I can access it from within the code, or just push your proposed changes.

PR #89, opened 2023-08-09
remove tag from Exposures

emanuel-schmid:
Changes proposed in this PR:
- Remove all tag occurrences from exposures classes

This PR fixes #

### PR Author Checklist

### PR Reviewer Checklist


climada_petals/entity/exposures/black_marble.py
```diff
@@ -116,7 +116,6 @@ def set_countries(self, countries, ref_year=2016, res_km=None, from_hr=None,
             data=Exposures.concat(bkmrbl_list).gdf,
             crs=DEF_CRS,
             ref_year=ref_year,
-            tag=Tag(file_name=fn_nl, description=description),
```
chahank, line 119: Also in line 36 above we can remove the import.

climada_petals/entity/exposures/gdp_asset.py
```diff
@@ -86,7 +86,6 @@ def set_countries(self, countries=[], reg=[], ref_year=2000,
             self,
             data=Exposures.concat(gdp2a_list).gdf,
             ref_year=ref_year,
-            tag=Tag(description=description),
```
chahank, line 89: The tag import in line 31 can be remove I think.

requirements/env_docs.yml
```diff
@@ -8,7 +8,6 @@ dependencies:
   - ipython
   - ipykernel
   - mccabe>=0.6
-  - nb_conda
```
chahank, line 11: This is a bit unrelated to this PR imho. Why is this necessary for the tag removal?

chahank (changes requested):
Small comments only, otherwise good, thanks!

requirements/env_docs.yml
emanuel-schmid: Right, this is only for making readthedocs run again. It's now in develop as well.

climada_petals/entity/exposures/gdp_asset.py
emanuel-schmid: Thanks for pointing this out!

emanuel-schmid:
Eventually it turned out to be a bit more complicated than hoped for. Three exposures classes in petals actually do store information about how the dataset was constructed in the tag's description. This information is shown in the exposures plots and to me doesn't see to be completely useless. Furthermore the datasets from the API seem to have contain information in the tags that I'm not sure is available otherwise.
So I decided to keep the description - for the time being - as an attribute, not just for `LitPop` but for all `Exposures` classes.

Besides I also kept the raw data files for `BlackMarble` and `SpamAgrar`, not as `file_name` but as `nightlight_file` and `spam_file` respectively, because the identification of the suitable file is part of the creation process and not provided by the user.

🤷

chahank:
I think we should get rid of these description elements unless we are able to make a stable and useful solution. The information in the plots are almost always not necessary. Concerning the data API, I would say it is a bad practice if relevant information is hidden in the attribute of a file instead of being part of the metadata of the file in the database. Thus, I would again argue for removing the description rather than keeping it.

emanuel-schmid:
As far as I'm concerned I agree, but whether something is useful or not is in the eye of the beholder.
You say the information in the plots is _almost always_ not necessary - sounds to me like _exceptionally_ they may benecessary.

Concerning the data API: fully agreed that is bad practice. But in case it was ever applied, we would need to replace the concerned datasets in a rather elaborate procedure that involves the `api_client.Client` from climada 3.x - or rebuild them from scratch.

On the other side: having a `description` attribute in the `Exposures` class looks quite harmless to me. There is an arbitrary string in the class obviously containing a description of the object - so what? And if it shows up in plots - all the better!
Is it useful? s.a. Is it stable? Stable enough I'd say.

climada_petals/entity/exposures/black_marble.py
```diff
@@ -102,22 +105,23 @@ def set_countries(self, countries, ref_year=2016, res_km=None, from_hr=None,
...
-        Exposures.__init__(
+        BlackMarble.__init__(
             self,
             data=Exposures.concat(bkmrbl_list).gdf,
             crs=DEF_CRS,
             ref_year=ref_year,
-            tag=Tag(file_name=fn_nl, description=description),
-            value_unit='USD'
+            value_unit='USD',
+            description="\n".join(descr_lines),
```
peanutfun, line 123: How does this work? The `description` is not added to the `Exposures._metadata`, so I assume it is *not* added as attribute but passed on to the `GeoDataFrame` constructor? How does that one deal with a string as argument? I am quite certain something goes wrong here.

climada_petals/entity/exposures/crop_production.py
```diff
@@ -363,9 +362,9 @@ def from_isimip_netcdf(cls, input_dir=None, filename=None, hist_mean=None,
             value_tmp = np.squeeze(area_crop[irr_val]*hist_mean_dict[irr_val][idx_mean])
             value_tmp = np.nan_to_num(value_tmp) # replace NaN by 0.0
             exp.gdf['value'] += value_tmp
-        exp.tag = Tag(description="Crop production exposure from ISIMIP"
-                                  f" {CROP_NAME[crop]['print']} {irr}"
-                                  f" {yearrange[0]} {yearrange[-1]}")
+        exp.description=("Crop production exposure from ISIMIP"
+                        f" {CROP_NAME[crop]['print']} {irr}"
+                        f" {yearrange[0]} {yearrange[-1]}")
```
peanutfun, line 367: Similar topic: I find it bad practise to set an attribute that is not defined in the constructor. (I know that `Exposures` combines several bad practices when it comes to constructors but we don't have to make it worse)

The `description` attribute is also never documented anywhere.

CHANGELOG.md
```diff
@@ -1,23 +1,25 @@
...
 - shapely `1.8` -> `2.0` ([#80](https://github.com/CLIMADA-project/climada_petals/pull/80))
 
 ### Added
 
 ### Changed
 
 - refactored `climada_petals.river_flood.RiverFlood.from_nc`, removing calls to `set_raster` ([#80](https://github.com/CLIMADA-project/climada_petals/pull/80))
+- several adaptations for the removal of the `Tag` class from the CLIMADA core package:
+  [#88](https://github.com/CLIMADA-project/climada_petals/pull/88)
+  [#89](https://github.com/CLIMADA-project/climada_petals/pull/89)
```
peanutfun, line 22:
```suggestion
- Replace `tag` attribute with string `description` in classes derived from `Exposure` [#89](https://github.com/CLIMADA-project/climada_petals/pull/89)
```

peanutfun (changes requested):
We should make `description` a "regular" attribute of the exposure classes that use it, and document it accordingly. Other than that, all looks fine.

Thank you for improving the tests by using better asserts than `assertTrue`, this is much better! 🙌

climada_petals/entity/exposures/black_marble.py
emanuel-schmid: Good question! I cheated there and sneaked `description` in as an attribute of the base class: https://github.com/CLIMADA-project/climada_python/commit/438be81def2488da58319823bf8a8b11d5a21616
Kind of had to otherwise the two repositories would have been in a catch 22 state. 😏

peanutfun: I see 👍

PR #100, opened 2023-11-24
Feature/coastal flooding

aleeciu:
Changes proposed in this PR:
- this PR loads coastal flood data from the [Aqueduct project](https://www.wri.org/data/aqueduct-floods-hazard-maps) and create a CLIMADA hazard instance from them.
- since it mostly uses established CLIMADA functions, test aim at testing the validity of (a subset of) the files' urls.

### PR Author Checklist

### PR Reviewer Checklist


climada_petals/hazard/coastal_flood.py
```diff
@@ -0,0 +1,127 @@
...
+        https://www.wri.org/data/aqueduct-floods-hazard-maps
+    """
+
+    def __init__(self, **kwargs):
+        """Empty constructor"""
+
+        Hazard.__init__(self, HAZ_TYPE, **kwargs)
+
+    @classmethod
+    def from_aqueduct_tif(cls,
```
chahank, line 30: In this case I think it makes much more sense to remove the `from` method and put all the functionality in the init. There seems to be no other purpose than to load data (initialize the object).

climada_petals/hazard/test/test_coastal_flood.py
```diff
@@ -0,0 +1,73 @@
...
+PERCENTILES = ['5', '50', '95']
+
+def rcp_name(rcp):
+    return f"rcp{rcp[0]}p{rcp[1]}" if rcp in ['45', '85'] else rcp
+
+def rp_name(return_period):
+    return f"{str(return_period).zfill(4)}"
+
+def perc_name(percentile):
+    return f"0_perc_{percentile.zfill(2)}" if percentile in ['05', '50'] else '0'
```
chahank, line 40: Why not define these methods in the main class? Would this not make more sense?

chahank (changes requested):
Cool addition! 

@emanuel-schmid : can you please just check how we should handle the external resources (here a weblink)? In particular, should we do something for the case that the link changes of the data is no longer available?

climada_petals/hazard/test/test_coastal_flood.py
aleeciu: I removed the use of these functions altogether

climada_petals/hazard/coastal_flood.py
aleeciu: mhh, as of now maybe yes, but it's likely that the CoastalFlood class will be populated with more data in the future; same as it's happening with the RiverFlood class where we now have ISIMIP and Aqueduct (https://github.com/CLIMADA-project/climada_petals/blob/feature/river_flooding_aqueduct/climada_petals/hazard/river_flood.py) so I would tend to keep it like this.

chahank:
Any update on this: @emanuel-schmid : can you please just check how we should handle the external resources (here a weblink)?

climada_petals/hazard/coastal_flood.py
```diff
@@ -0,0 +1,134 @@
...
+
+from climada.hazard import Hazard
+from climada.util import files_handler as u_fh
+import climada.util.coordinates as u_coord
+from climada import CONFIG
+
+AQUEDUCT_SOURCE_LINK = CONFIG.hazard.coastal_flood.resources.aqueduct.str()
+# FIX THIS PATH - YOU SHOULD SAVE UNDER HAZARD, NOW IT CREATES A FOLDER CALLED COASTALFLOOD
+# AT THE SAME LEVEL AS HAZARD. RIVERINE AQUEDUCT IS SAVED THERE TOO, SO FIX THAT TOO
+# (SEE CLIMADA DATA LOCAL FOLDER)
```
emanuel-schmid, line 13: Is this comment still valid? Actually the path ~/climada/data/hazard is where the files from the data api land. There are some automatic features on this directory executed by the api client: caching, making sure that the version is always the newest and purging. So, strictly speaking, the data files that are not handled by the climada api client _should_ not necessarily be saved there. Any other directory on the same level actually seems better suited.

chahank:
Any update on this @aleeciu and @emanuel-schmid  ?

aleeciu:
Hi @chahank , meanwhile @emanuel-schmid and I actually made these data accessible via the API. So I wonder if there is any point in completing this PR in the first place. All it does is loading data that are now accessible via the API.

bguillod:
> Hi @chahank , meanwhile @emanuel-schmid and I actually made these data accessible via the API. So I wonder if there is any point in completing this PR in the first place. All it does is loading data that are now accessible via the API.

What is your take on this @chahank @emanuel-schmid ?

emanuel-schmid:
Imho, just because the resulting data is available elsewhere doesn't mean that the code is obsolete.
Quite in the opposite: a reference on how the data has been created would be a very good idea.
The data-type on the API does not give much information away about that:

image

chahank:
@bguillod @aleeciu : any plans to finish this PR?

climada_petals/hazard/coastal_flood.py
aleeciu: removed

aleeciu:
Given that we clarified the issue with the config file, from my end this PR is complete. As for #153 I seem to be getting test errors about wind calculation. I don't know where this comes from.

emanuel-schmid:
Unfortunately, there are no tests at all for the `CoastalFlood` class. The tests in `test_coastal_flood` just randomly test whether the original resource would be available for `CoastalFlood` in case it wanted to fetch them.
This is nice to know and would be valuable as a test for the [petals_data_api](https://ied-wcr-jenkins.ethz.ch/job/petals_data_api/) workflow or even in the [integration test](https://ied-wcr-jenkins.ethz.ch/job/petals_ci_night/). As a unit test it is none. 😄 

Since the `CoastalFlood` class' main purpose is downloading files and turning them into hazard objects, which takes too long for unittests, it provides a perfect use case for _mocking_. I reckon, by mocking `u_fh.download_file` and `Hazard.from_raster`, one could rapidly confirm the correct behavior of `CoastalFlood`.

aleeciu:
Hi @emanuel-schmid, from my point of view, since the new class makes use of existing and tested functions, the only thing I had to test was that the files actually exist and can be downloaded. This was also discussed and agreed with @chahank and @peanutfun but like 2 years ago maybe, so they might not remember :-)

Anyway, I am happy to add a test using _mocking_, if I knew what that is in the first place :-) (well, from its name I can guess what's the use of it, but still). So if you have example or if it's even fast to do it on your end, that would be highly appreciated. I guess I could then implement the same for #108

chahank:
Great!

Here is the link to the documentation for mocking: https://docs.python.org/3/library/unittest.mock.html

You can also find some mocking in the tests for impact calc: https://github.com/CLIMADA-project/climada_python/blob/main/climada/engine/test/test_impact_calc.py

bguillod:
Thanks @emanuel-schmid for merging develop back into this. From our perspective this one is ready, so we just need @chahank's approval. Once this is done, develop can be again merged into #108 and I will review that other one.

emanuel-schmid:
Thanks @bguillod. One last thing - can you add something in the change log?

bguillod:
@emanuel-schmid Now @jlehtomaa has just added a changelog entry for this and #108 in #108 - can we merge this one as is, such that then we can merge develop into #108 and I'll review that one once this is done?

emanuel-schmid:
@bguillod: hm. 🤔 yes we could, but imho, the changelog entry in #108 is a bit too general and vague. I'd expect somthing like

```md
## Added
- module `climada.hazard.coastal_flood` [#100](https://github.com/CLIMADA-project/climada_petals/pull/100)
```

bguillod:
> @bguillod, @jlehtomaa: hm. 🤔 yes we could, but imho, the changelog entry in #108 is a bit too general and vague. I'd expect somthing like
> 
> ```
> ## Added
> - module `climada_petals.hazard.coastal_flood` [#100](https://github.com/CLIMADA-project/climada_petals/pull/100)
> ```

@emanuel-schmid Thanks, I've done this one already here - good like this?
Once merged I'll adjust #108 accordingly

bguillod:
Thanks @emanuel-schmid for the last fixes - are we good to merge this one now?

PR #103, opened 2023-11-27
Outsource OpenStreetMap data dump access to osm-flex

Evelyn-M:
Changes proposed in this PR:
- Make `osm-flex` a dependency, and re-introduce `overpy` as dependency (else accessing overpass API fails)
- Delete all CLIMADA-internal code which downloaded, clipped and parsed from osm.pbf files
- Update openstreetmap tutorial and tests

This PR fixes #

### PR Author Checklist

### PR Reviewer Checklist


leonie-villiger (changes requested):
Thank you for updating the `openstreetmap` (now `osm_dataloader`) module! I could follow your actions and learned a lot when executing the tutorial 😃   

There are *three issues" that must be resolved before merging:
* update the module name in the tutorial (see detailed comment there)
* similarly, update the name of the testfile (still named after the old module name)
* ensure that the `climada_env` can be update using `conda update` (currently the `osm-flex` is not found and can only be installed using `pip install`).

Additionally, I suggested some minor changes. 

**Important**: Due to my lack of expertise about defining classes (i.e. `OSMApiQuery`), I suggest that a second reviewer has a look at the files
- `climada_petals/entity/exposures/osm_dataloader.py`
- `climada_petals/entity/exposures/test/test_openstreetmap.py`
I looked through it, but didn't think through all the functionalities.

requirements/env_climada.yml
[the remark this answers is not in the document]
Evelyn-M: I believe `osm-flex` can only be installed using the `pip install` command. @chahank or @peanutfun, can you help me out with how to specify this in the env-file? Or if this is not the case, do you know why specifying osm-flex like - `- osm-flex>=1.1.1` does not work?

climada_petals/entity/exposures/test/test_openstreetmap.py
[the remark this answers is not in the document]
Evelyn-M: Good point. renamed.

doc/tutorial/climada_exposures_openstreetmap.ipynb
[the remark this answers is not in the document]
Evelyn-M: Thanks, very useful comments. Incorporated them accordingly.
Last points to 
2.1:  that is content of and explained in the lines_polys tutorial, hence I would refrain from duplicating explanations here. I did add a comment on where to find a detailed interpolation tutorial though.

2.2: True, in the end I thought the following steps are really similar to any CLIMADA calculation and the lines_poly tutorial, hence this should be no problem for the user to continue from there, plus it's really not OSM specific anymore.

climada_petals/entity/exposures/test/test_openstreetmap.py
```diff
@@ -131,73 +51,71 @@ def test_area_to_queryformat(self):
...
+    
     def test_osm_geoms_to_gis(self):
         """test methods of OSMApiQuery" """
         pass
-
+    
     def skip_test_get_data_overpass(self):
-        """test methods of OSMApiQuery"
+        """test methods of OSMApiQuery" 
         skipping: test causes a segfault on Jenkins"""
```
peanutfun, line 104: Is this information still up-to-date? Do we know what causes the segfault?

climada_petals/entity/exposures/osm_dataloader.py
```diff
@@ -0,0 +1,346 @@
...
+            return (ymin, xmin, ymax, xmax)
+
+        if isinstance(area,shapely.geometry.Polygon):
+            lon, lat = area.exterior.coords.xy
+            lat_lon_str = " ".join([str(y)+" "+str(x) for y, x in zip(lat, lon)])
+            return f'(poly:"{lat_lon_str}")'
+    
+        return None
+
+    def _insistent_osm_api_query(self, query_clause, read_chunk_size=100000,
```
peanutfun, line 62: Jenkins has a point: This method uses nothing in `self`, so it could become a "free" function instead

climada_petals/entity/exposures/osm_dataloader.py
```diff
@@ -0,0 +1,346 @@
...
+LOGGER = logging.getLogger(__name__)
+
+class OSMApiQuery:
+    """
+    Queries features directly via the overpass turbo API.
+
+    area: tuple (xmin, ymin, xmax, ymax), list [xmin, ymin, xmax, ymax]
+        or shapely.geometry.Polygon
+    query: str
+        must be of format '["key"]' or '["key"="value"]', etc.
```
peanutfun, line 38: This information is outdated. There is no `query` anymore, but `condition`. Notice also that the *attribute* `area` is either a tuple or a string, whereas the *parameter* `area` to the `__init__` can be a sequence or a polygon. Please make this distinction clear.

climada_petals/entity/exposures/osm_dataloader.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def _area_to_queryformat(self, area):
+        """
+        reformat lat/lon info as in OSM convention, meaning
+        bbox: (S,W,N,E) instead of (xmin, ymin, xmax, ymax)
+        Points: lat / lon instead of (x,y)
+        """
+        if isinstance(area,(tuple, list)):
+            xmin, ymin, xmax, ymax = area
+            return (ymin, xmin, ymax, xmax)
```
peanutfun, line 53: This is very confusing. What happens when you enter the `area` in the right order? Then the resulting order is wrong. One quick idea I have is to create two `classmethods` that create the object from different data types:

```python
def __init__(self, area, condition):
    self.area = area
    self.condition = condition

@classmethod
def from_bounding_box(cls, bbox, condition):
    xmin, ymin, xmax, ymax = bbox  # Maybe need to make sure that bbox is what you expect it is?
    return cls((ymin, xmin, ymax, xmax), condition)

@classmethod
def from_polygon(cls, polygon, condition):
    lon, lat = area.exterior.coords.xy
    lat_lon_str = " ".join([str(y)+" "+str(x) for y, x in zip(lat, lon)])
    return cls(area=f'(poly:"{lat_lon_str}")', condition=condition)
```

climada_petals/entity/exposures/osm_dataloader.py
```diff
@@ -0,0 +1,346 @@
...
+                    shapely.ops.linemerge(shapely.geometry.MultiLineString(lines)))
+            else:
+                poly = shapely.geometry.Polygon([])
+
+            # step 3: combine to one multipoly
+            multipoly = shapely.ops.unary_union([outer_mp - inner_mp, poly])
+            data_geom.append(multipoly)
+
+            if multipoly.area == 0:
+                LOGGER.info('Empty geometry encountered.')
```
peanutfun, line 171: Is this important information? What is the user to do about that? If there is nothing one can do and the code handles this case well, then it should be a `debug` information

climada_petals/entity/exposures/osm_dataloader.py
```diff
@@ -0,0 +1,346 @@
...
+
+        gdf_rels =  gpd.GeoDataFrame(
+            data={'osm_id': data_id,'geometry': data_geom, 'tags':data_tags})
```
peanutfun, line 174: This raises a warning. GeoDataFrames must be initialized with explicit `geometry` parameter. Also, the CRS is missing.

climada_petals/entity/exposures/osm_dataloader.py
```diff
@@ -0,0 +1,346 @@
...
+                data_geom.append(shapely.geometry.LineString(node_lat_lons))
+                data_id.append(way.id)
+                data_tags.append(way.tags)
+
+            if closed_lines_are_polys:
+                data_geom = [shapely.geometry.Polygon(way) if way.is_closed
+                             else way for way in data_geom]
+
+        gdf_ways = gpd.GeoDataFrame(
+            data={'osm_id': data_id,'geometry': data_geom, 'tags':data_tags})
```
peanutfun, line 230: Same as above, please initialize correctly.

climada_petals/entity/exposures/osm_dataloader.py
```diff
@@ -0,0 +1,346 @@
...
+            Not "available" anymore for further constructions
+        gdf_rels : gpd.GeoDataFrame
+            gdf with relations that were assembled from result object
+        """
+
+        nodes_taken = []
+        ways_taken = []
+        data_geom = []
+        data_id = []
+        data_tags = []
```
peanutfun, line 125: Having multiple lists and appending to them is dangerous, because you might end up with lists of different lengths. Appending is also quite inefficient when objects or lists get large. Could you maybe make the loop that appends on these lists a separate function, so the lists can be assigned to the results immediately through a list comprehension?

climada_petals/entity/exposures/osm_dataloader.py
```diff
@@ -0,0 +1,346 @@
...
+
+        for relation in result.relations:
+            data_tags.append(relation.tags)
+            data_id.append(relation.id)
+            roles = []
+            relationways = []
+
+            for way in relation.members:
+                relationways.append(way.ref)
+                roles.append(way.role)
```
peanutfun, line 135: Clearer this way. Or is iterating over `relation.members` very costly?
```suggestion
            roles = [way.ref for way in relation.members]
            relationways = [way.role for way in relation.members]
```
You could also create the `gdf_polys` after `relationways`, but before `roles`, then assign the `role` column as you do below directly. This way, you drop the superfluous `role` variable. (Jenkins also complains about too many variables). Example:
```python
relationways = [way.role for way in relation.members]
nodes_taken_mp, gdf_polys = self._assemble_from_ways(
    result, relationways, closed_lines_are_polys=True
)
gdf_polys['role'] = [way.ref for way in relation.members]
```

climada_petals/entity/exposures/osm_dataloader.py
```diff
@@ -0,0 +1,346 @@
...
+        data_tags = []
+
+        for node in result.nodes:
+            if node.id in nodes_avail:
+                data_geom.append(shapely.geometry.Point(node.lat, node.lon))
+                data_id.append(node.id)
+                data_tags.append(node.tags)
+
+        gdf_nodes = gpd.GeoDataFrame(
+            data={'osm_id': data_id,'geometry': data_geom, 'tags':data_tags})
```
peanutfun, line 263: Same as above, please initialize correctly.

climada_petals/entity/exposures/osm_dataloader.py
```diff
@@ -0,0 +1,346 @@
...
+        closed_lines_are_polys : bool
+                whether closed lines are polygons. Default is True
+
+        Returns
+        ------
+        gdf_results : gpd.GeoDataFrame
+            Result-gdf from the overpass query.
+        """
+        gdf_results = gdf_ways = gdf_nodes = gpd.GeoDataFrame(
+            columns=['osm_id','geometry','tags'])
```
peanutfun, line 291: Similar to above. CRS is missing, geometry is not set explicitly. In general, I would say it does not help to initialize an empty GeoDataFrame for this reason. Simply take the ones returned by the other methods and concatenate them, instead of appending to the empty one.

climada_petals/entity/exposures/osm_dataloader.py
```diff
@@ -0,0 +1,346 @@
...
+
+    def get_data_overpass(self, closed_lines_are_polys=True):
+        """
+        wrapper for all helper funcs to get & assemble data
+        """
+
+        query_clause = self._overpass_query_string()
+        result = self._insistent_osm_api_query(query_clause)
+        gdf_result = self._assemble_results(result, closed_lines_are_polys)
+        gdf_result['geometry'] = self._osm_geoms_to_gis(gdf_result)
```
peanutfun, line 343: Geometry should be set explicitly, see https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.set_geometry.html
```suggestion
        gdf_result = gdf_result.set_geometry(self._osm_geoms_to_gis(gdf_result))
```

climada_petals/entity/exposures/osm_dataloader.py
```diff
@@ -0,0 +1,346 @@
...
+        if len(nodes_avail) > 0:
+            gdf_nodes = self._assemble_from_nodes(result, nodes_avail)
+            gdf_results =  gdf_results.append(gdf_nodes)
+
+        if len(result.nodes) == 0:
+            LOGGER.warning('empty result gdf returned.')
+
+        return gdf_results.reset_index(drop=True)
+
+    def _osm_geoms_to_gis(self, gdf):
```
peanutfun, line 316: This method should only take the geometry as parameter, because it does not access the rest of the dataframe. Then it can be called like
```python
geo_new = _osm_geoms_to_gis(gdf.geometry)
```

climada_petals/entity/exposures/test/test_openstreetmap.py
```diff
@@ -122,7 +42,7 @@ def test_osmqpiquery(self):
 
     def test_area_to_queryformat(self):
         """ test methods of OSMApiQuery"""
-
+
```
peanutfun, line 45: A lot of unnecessary whitespace added in this file. Please remove. Jenkins does not complain because our pylint does not parse the tests.
```suggestion

```

climada_petals/entity/exposures/test/test_openstreetmap.py
```diff
@@ -131,73 +51,71 @@ def test_area_to_queryformat(self):
...
-
+    
 # Execute Tests
 if __name__ == "__main__":
-    TESTS = unittest.TestLoader().loadTestsFromTestCase(TestOSMRaw)
-    TESTS.addTests(unittest.TestLoader().loadTestsFromTestCase(TestOSMFileQuery))
-    TESTS.addTests(unittest.TestLoader().loadTestsFromTestCase(TestOSMApiQuery))
-    unittest.TextTestRunner(verbosity=2).run(TESTS)
+    TESTS = unittest.TestLoader().loadTestsFromTestCase(TestOSMApiQuery)
+    unittest.TextTestRunner(verbosity=2).run(TESTS)
```
peanutfun, line 121: Missing newline at end of file
```suggestion
    unittest.TextTestRunner(verbosity=2).run(TESTS)

```

requirements/env_climada.yml
```diff
@@ -3,5 +3,8 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - scikit-image>=0.22
+  - overpy
   - pymrio>=0.5
+  - scikit-image>=0.22
+  #- pip:
+  #    - osm-flex>=1.1.1
```
peanutfun, line 10: Make sure you add `osm-flex` to the requirements in `setup.py`
```suggestion
```

climada_petals/entity/exposures/test/test_openstreetmap.py
```diff
@@ -131,73 +51,71 @@ def test_area_to_queryformat(self):
...
+        
         self.assertEqual(
             q1, '[out:json][timeout:180];(nwr["building"](47.36826, 8.5327506, 47.376877, 8.5486078);(._;>;););out;')
         self.assertEqual(
             q2, '[out:json][timeout:180];(nwr["amenity"="place_of_worship"](poly:"47.36826 8.5327506 47.376877 8.5486078 47.39 8.5486078 47.36826 8.5327506");(._;>;););out;')
-
+    
     def test_assemble_from_relations(self):
         """test methods of OSMApiQuery" """
         pass
```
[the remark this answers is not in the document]
Evelyn-M: They should be filled 👼

climada_petals/entity/exposures/test/test_openstreetmap.py
Evelyn-M: not sure, i'll try not skipping it and see what happens

climada_petals/entity/exposures/osm_dataloader.py
Evelyn-M: true. since there were many functions organised in 4 classes in this module initially, it made sense to put this in its parent class for organisational reasons. there are still a few more of these. will make sense to restructure / entirely remove the OSMApiQuery class in the future.

climada_petals/entity/exposures/osm_dataloader.py
Evelyn-M: updated it accordingly.

climada_petals/entity/exposures/test/test_openstreetmap.py
Evelyn-M: not failing anymore, hence not skipping htem :)

peanutfun:
@Evelyn-M What's the rush? Several of my comments have not been addressed, and there is quite a number of linter issues one could still resolve, at least two of which have elevated severity. @leonie-villiger and I both requested changes which means that we have to have another round that approves the new version. If you need to hit a deadline, please say so earlier.

climada_petals/entity/exposures/test/test_openstreetmap.py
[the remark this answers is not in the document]
peanutfun: Who's doing that when?

Evelyn-M:
Sorry, there were a few hidden comments that got lost!  Luckily only 2 or 3, and they're anyways part of the APIQuery part, which was actually not intended to be changed as part of this pull request (I still made quite a few changes upon your suggestions, but feel free to open a separate PR for further amendements. They're all good points, so I'll copy them over to the respective GitHub issue.) Leonie did re-review and approve, and I have no work-time left on this (sorry for the rush this time, was a pragmatic decision :) ).

PR #111, opened 2024-02-06
Feature/add missing jrc flood impact functions

aleeciu:
This PR adds the missing flood impact fucntions from the JRC publications. The current implementation only uses the "residential buildings" functions, and this PR adds those from the five more categories (or sectors as tjey are called here) available in the JRC report, i.e., commercial buildings, industrial buildings, transport, infrastructure and agriculture. To do that, I had to slightly adjust the code as one can no longer only choose the region, but both the region and a sector.


climada_petals/entity/impact_funcs/test/test_fl.py
```diff
@@ -35,116 +35,114 @@ def test_flood_imp_func_set(self):
 
     def test_region_Africa(self):
```
chahank, line 36: Would it not be simpler to have tests that loop through the allowed parameters and all methods? This class just stores data, so this is what should be tested. Also, a test for when wrong parameters are used would be good. This makes the test more complete, and also shorter.

chahank:
@aleeciu : could you please provide a timeline to finish this PR?

aleeciu:
Apologies I replied quite timely in fact, but I did not realize I had to "send the review" in order for my comment to be visible. Anyway, here you really find me in line with comments on test in https://github.com/CLIMADA-project/climada_python/pull/734 : I am surprised about the request because the logic I followed is just the one already in the module so I find writing better tests very very important but simply beyond scope, and increasing scope is how one never gets things done :-)

aleeciu:
I added tests looping through all possible combinations while catching wrong combinations. I also added test for wrong value assignments. In my view this is all it was asked in the review. Please let me know.

climada_petals/entity/impact_funcs/test/test_fl.py
chahank: That would be a good idea. The code base is constantly improving, and thus copying older, less mature code is suboptimal going forward.

climada_petals/entity/impact_funcs/river_flood.py
```diff
@@ -55,154 +53,452 @@ def __init__(self):
...
+        impf = cls()
+        impf.name = f"Flood {region} JRC {sector.capitalize()} noPAA"
+        impf.continent = f"{region}"
+        impf.id = impf_id
+        impf.mdd = impf_values
+        impf.mdr = impf_values
+        impf.intensity = np.array([0., 0.05, 0.5, 1., 1.5, 2., 3., 4., 5., 6., 12.])
         impf.paa = np.ones(len(impf.intensity))
+
         return impf
```
chahank, line 123: Please use the init of the class properly. The init variables are:

```
        id: Union[str, int] = "",
        intensity: Optional[np.ndarray] = None,
        mdd: Optional[np.ndarray] = None,
        paa: Optional[np.ndarray] = None,
        continent: str = ''
        name: str = ""

So, the code should be 

return cls(
    id = impf_id,
    mdd = impf_values,
    paa = np.ones(len(intensity)),
    intensity = intensity,
    name = name,
    continent=continent
)
```

climada_petals/entity/impact_funcs/river_flood.py
```diff
@@ -55,154 +53,452 @@ def __init__(self):
...
+
+        elif sector == 'transport':
+            impf_values, impf_id = from_jrc_impf_transport(region)
+
+        elif sector == 'infrastructure':
+            impf_values, impf_id = from_jrc_impf_infrastructure(region)
+
+        elif sector == 'agriculture':
+            impf_values, impf_id = from_jrc_impf_agriculture(region)
+
```
chahank, line 110: Long chains of `if`  and `else` statements are to be avoided. For future work please use a dataclass or similar.

climada_petals/entity/impact_funcs/river_flood.py
```diff
@@ -55,154 +53,452 @@ def __init__(self):
...
 
-def flood_imp_func_set():
-    """Builds impact function set for river flood, using standard files"""
+    return impf_values, impf_id
+
+def flood_imp_func_set(sector="residential"):
+    """Builds impact function set for river flood, using standard files. By default, it reads
+    functions for the residential sector"""
 
     impf_set = ImpactFuncSet()
```
chahank, line 475: Please use a proper initialization. Something like:

impf_Set = ImapctFuncSet([
   impf_africa,
   impf_asia,
   ...
])

climada_petals/entity/impact_funcs/river_flood.py
```diff
@@ -55,154 +53,452 @@ def __init__(self):
...
     def set_RF_Impf_SouthAmerica(self, *args, **kwargs):
-        """This function is deprecated, use ImpfRiverFlood.from_region instead."""
+        """This function is deprecated, use ImpfRiverFlood.from_region_sector instead."""
         LOGGER.warning("The use of ImpfRiverFlood.set_RF_Impf_SouthAmerica is deprecated."
-                       "Use ImpfRiverFlood.from_region instead.")
-        self.__dict__ = ImpfRiverFlood.from_region("SouthAmerica", *args, **kwargs).__dict__
+                       "Use ImpfRiverFlood.from_region_sector instead.")
+        self.__dict__ = ImpfRiverFlood.from_jrc_region_sector("SouthAmerica", *args, **kwargs).__dict__
+
+def from_jrc_impf_residential(region):
```
chahank, line 161: This would probably have been much better solved with a dataclass. However, I am not sure it is worth the effort now. @emanuel-schmid what do you think?

climada_petals/entity/impact_funcs/river_flood.py
chahank: Also, ImpactFunc does not have an attribute `.mdr`. Please make sure to remove this.

chahank (changes requested):
Excellent! Please just update the code to use init methods properly. 

For future pull request it might be beneficial to write new quality code following the guidelines instead of copying legacy code. 

Please also update the CHANGELOG file so that these changes are recorded for the next release. 

Otherwise, concerning the current style of the methods with various `if/else statements`, I would appreciate confirmation from @emanuel-schmid that it is fine to leave it as is.

aleeciu:
> Excellent! Please just update the code to use init methods properly.
> 
> For future pull request it might be beneficial to write new quality code following the guidelines instead of copying legacy code.
> 
> Please also update the CHANGELOG file so that these changes are recorded for the next release.
> 
> Otherwise, concerning the current style of the methods with various `if/else statements`, I would appreciate confirmation from @emanuel-schmid that it is fine to leave it as is.

Thanks a lot. An update:

- I will change the method initiation
- Concerning legacy code and following guidelines: I must say I read the guidelines but, unless I am reading the wrong ones, they are much less descriptive than what you say they are :-)
- Waiting for @emanuel-schmid to give green light on current status or refactoring is needed

emanuel-schmid:
Actually no, I couldn't overcome myself and give green light. 😬
Instead I committed some beautifications. Please have a look @aleeciu. If you think it's still ok, you're welcome to merge.

climada_petals/entity/impact_funcs/river_flood.py
aleeciu: removed

climada_petals/entity/impact_funcs/river_flood.py
aleeciu: adjusted as suggested

emanuel-schmid:
@aleeciu the failing test is actually related to this pr: before there were 86 lines covered and 16 missed, now there are 79 lines covered and 15 missed, which is a slightly lower ratio. Just because you made the code denser, so to say. 🤷 
We can safely ignore it.

emanuel-schmid:
@aleciu can you please fix the failing integration test: `climada_petals.test.test_flood_integr.TestRiverFlood.test_full_impact` and submit a new PR?

One problem is the DEMO exposures has `3` for each `impf_RF` value, while the ids from `flood_imp_func_set()` has id's `[11, 21, 31, 41, 51, 61]`.

The other problem is that the result differs significantlly:
```
test_flood_integr.py", line 80, in test_full_impact
    self.assertAlmostEqual(imp.at_event[0], 226839.72426476143)
AssertionError: 188191.9029567669 != 226839.72426476143 within 7 places (38647.82130799454 difference)
```

aleeciu:
I opened a new PR https://github.com/CLIMADA-project/climada_petals/pull/153

PR #122, opened 2024-04-11
Feature/centroids as gdf

emanuel-schmid:
Changes proposed in this PR:
- Adaptations to the centroids as gdf refactoring (climada 5.0.0-dev)

This PR fixes #

### PR Author Checklist

### PR Reviewer Checklist


peanutfun:
@emanuel-schmid The code coverage test fails because the coverage of modified lines is too low (49%, target: 60%). Is there something we want to do about this? Do we want to add some more testing to meet the target? If we simply ignore it, the pipeline for `develop` might succeed after merging, because the overall quality gate might still met and there will be no check of modified files only.

emanuel-schmid:
Yes, ignore. In this particular PR I don't see a point in adding tests - it's about making necessary adaptations to be compatible with core develop and not about improving anything at all.

I'm actually inclined to reconfigure the coverage restrictions. We could try the `MODIFIED_LINES_DELTA` baseline. I wonder though what we should set as `THRESHOLD` then - 100%?. And I wonder how it affects novel modules.
Also - what about plotting and I/O methods which are supposed to be in the integration test?
Perhaps we shouldn't make coverage cause failures at all.

peanutfun:
Alright, I should have the review ready by Wednesday

Makefile
```diff
@@ -16,15 +16,15 @@ help:  ## Use one of the following instructions:
 
 .PHONY : lint
 lint : ## Static code analysis with Pylint
-	pylint -ry climada_petals > pylint.log || true
+	python -m pylint -ry climada_petals > pylint.log || true
 
 .PHONY : unit_test
 unit_test : ## Unit tests execution with coverage and xml reports
-	pytest $(PYTEST_ARGS) --ignore=climada_petals/test climada_petals/
+	python -m pytest $(PYTEST_ARGS) --ignore=climada_petals/test climada_petals/
```
peanutfun, line 23: I see you updated the Jenkins scripts. Why was this addition necessary? `python -m pytest` as opposed to `pytest` adds the current directory to the `sys.path`. This enables testing local source code of modules which are not installed. At the same time, it makes it less clear what code exactly is tested. I recommend making sure that Petals is *installed* (possibly from sources with `pip install -e`), and then removing `python -m`, to make sure that the installed version is tested. See the [GitHub Actions](https://github.com/CLIMADA-project/climada_petals/blob/develop/.github/workflows/testing.yml) for reference, where `python -m` was not required so far

climada_petals/hazard/river_flood.py
```diff
@@ -137,22 +136,22 @@ def from_nc(cls, dph_path=None, frc_path=None, origin=False,
             if ISINatIDGrid:
 
                 dest_centroids = RiverFlood._select_exact_area(countries, reg)[0]
-                meta_centroids = copy.copy(dest_centroids)
-                meta_centroids.set_lat_lon_to_meta()
+                centroids_meta = dest_centroids.get_meta()
```
peanutfun, line 139: Seems like the linter is not up-to-date with the latest version of `climada_python`? It complains that `Centroids.get_meta` does not exist, but it does in `develop`. This might be related to the changes in the Jenkins scripts.

climada_petals/hazard/tc_rainfield.py
```diff
@@ -397,21 +397,13 @@ def from_tracks(
         TCRain
         """
         num_tracks = tracks.size
-        if centroids is None:
-            centroids = Centroids.from_base_grid(res_as=360, land=True)
-
-        if not centroids.coord.size:
-            centroids.set_meta_to_lat_lon()
-
```
peanutfun, line 405: One could use the `climada.util.constants.NATEARTH_CENTROIDS` and still keep the centroids argument optional here

climada_petals/hazard/test/test_wildfire.py
```diff
@@ -274,20 +274,23 @@ def test_centroids_pass(self):
...
+        self.assertEqual(centroids_meta['height'], 138)
+        self.assertAlmostEqual(centroids_meta['transform'][0], 0.006749460043196544)
+        self.assertAlmostEqual(centroids_meta['transform'][1], 0.0)
+        self.assertLessEqual(centroids_meta['transform'][2], firms.longitude.min())
+        self.assertAlmostEqual(centroids_meta['transform'][3], 0.0)
+        self.assertAlmostEqual(centroids_meta['transform'][4], -centroids_meta['transform'][0])
+        self.assertGreaterEqual(centroids_meta['transform'][5], firms.latitude.max())
+        self.assertLessEqual(firms.latitude.max(), centroids.total_bounds[3])
+        # TODO: check whether allowing for a tolerance is valid here
+        TOLERANCE = 0.005
```
peanutfun, line 288: How to check that? 🤔

peanutfun (changes requested):
Nicely done! Most changes look really straightforward. However, I am confused about the changes to the Makefile and the Jenkins scripts in this PR. Are they related to the centroids PR at all? There are two other minor points. See my comments.

Makefile
emanuel-schmid: The addition is necessary on Jenkins because all the branches use the same conda environment. `pip install -e` would interfere with other branches' tests.
For referring to an arbitrary climada core branch one can install that by creating a _venv_ environment on top of the _conda_ environment. (Which is about 10 times faster). But in such a layered environment `pytest` is taken from the underlying _conda_ sources and imports climada from there, while `python -m pytest` starts from _venv_ and imports climada from here.
This was done, e.g., for exactly this PR before #787 was merged.

GitHub Actions on the other hand build a new conda environment for every branch and commit. I'm afraid we don't have enough resources in terms of cpu or diskspace to do that on Jenkins.

climada_petals/hazard/river_flood.py
emanuel-schmid: Right. The develop environment of climada_petals points to core package before the refactoring. Opposed to the `unit_test` stage, the `lint` stage does not change to the venv environment on top. That's why.

Makefile
peanutfun: I am still confused why this is necessary here in Petals but not in Core. Doesn't that mean there is something off with the environment configuration?

Anyway, I see that this is too much hassle to change. However, I would like to keep the default invoking of `pytest` for users, which should work best with the current installation instructions. I found a way to insert the different command into the Makefile, and will push it shortly.

peanutfun: Ok sorry I did not realize you added new bash scripts for Jenkins, see my comments there

peanutfun (commented):
I somehow missed the new bash scripts for Jenkins, and I have some trouble understanding what is going on. Why is the setup so much different from the one in Core? I think we should make sure that both users *and* the pipeline can simply use the `Makefile` provided.

script/jenkins/branches/run_unittest.sh
```diff
@@ -0,0 +1,22 @@
...
+CORENV=~/jobs/petals_branches/core_env
+BRANCH=`git name-rev --name-only HEAD | cut -f 3- -d /`
+if [ -f $CORENV/$BRANCH ]; then
+    python -m venv --system-site-packages tvenv
+    source tvenv/bin/activate
+
+    pip install -e `cat $CORENV/$BRANCH`
+fi
+
+make unit_test PYTEST_CMD="python -m pytest"
```
peanutfun, line 17: @emanuel-schmid This is how we can invoke a different command in the Makefile when executing it via Jenkins, see the changes in the Makefile as well

climada_petals/hazard/tc_rainfield.py
emanuel-schmid: gave it a try but found it unpractical as it covers the whole earth and causes computation downstream to last forever and consume all the memory.
Instead I took the boundaries from tracks and built a `Centroids` object spanning it. Now I wonder whether it should contain `on_land` coordinates only. But that could be done in a separate PR I suppose.

script/jenkins/branches/run_unittest.sh
emanuel-schmid: nice!

script/jenkins/branches/run_unittest.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+export PATH=$PATH:$CONDAPATH
+
+source activate petals_env
+rm -rf tests_xml/
+rm -rf coverage/
+
+CORENV=~/jobs/petals_branches/core_env
+BRANCH=`git name-rev --name-only HEAD | cut -f 3- -d /`
+if [ -f $CORENV/$BRANCH ]; then
```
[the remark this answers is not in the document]
emanuel-schmid: This all is about having specific core einvironments for petals branches (and vice versa)
It is unrelated to this PR now, but before #787 was merged into develop I needed a way to point to the sources from that branch. (The pointing itself is done using files on Jenkins by the name of the petals branch and content of the core branch: quick and dirty!)
We can undo all changes in Jenkinsfile and Makefile. They don't relate to this PR (anymore) and presumably we don't need them in the near future. 🤷

script/jenkins/branches/run_unittest.sh
```diff
@@ -0,0 +1,22 @@
...
+    python -m venv --system-site-packages tvenv
+    source tvenv/bin/activate
+
+    pip install -e `cat $CORENV/$BRANCH`
```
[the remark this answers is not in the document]
emanuel-schmid: It's not about expanding variables, it's about the content of that file.

climada_petals/hazard/test/test_tc_rainfield.py
```diff
@@ -106,7 +105,7 @@ def test_tcr(self):
         # This increases the results by more than 70% because the default value for saturation
         # specific humidity corresponds to a temperature of only ~267 K.
         tc_track.data[0]["t600"] = xr.full_like(tc_track.data[0]["central_pressure"], 275.0)
-        tc_haz = TCRain.from_tracks(tc_track, model="TCR", centroids=CENTR_TEST_BRB)
+        tc_haz = TCRain.from_tracks(tc_track, CENTR_TEST_BRB, model="TCR")
```
emanuel-schmid, line 108:
```suggestion
        tc_haz = TCRain.from_tracks(tc_track, model="TCR", centroids=CENTR_TEST_BRB)
```

climada_petals/hazard/test/test_tc_rainfield.py
```diff
@@ -95,7 +94,7 @@ def test_tcr(self):
         tc_track = TCTracks.from_processed_ibtracs_csv(TEST_TRACK)
         tc_track.equal_timestep()
 
-        tc_haz = TCRain.from_tracks(tc_track, model="TCR", centroids=CENTR_TEST_BRB)
+        tc_haz = TCRain.from_tracks(tc_track, CENTR_TEST_BRB, model="TCR")
```
emanuel-schmid, line 97:
```suggestion
        tc_haz = TCRain.from_tracks(tc_track, model="TCR", centroids=CENTR_TEST_BRB)
```

climada_petals/hazard/test/test_tc_rainfield.py
```diff
@@ -67,7 +66,7 @@ def test_set_one_pass(self):
         """Test from_tracks constructor with a single track."""
         tc_track = TCTracks.from_processed_ibtracs_csv(TEST_TRACK)
         tc_track.equal_timestep()
-        tc_haz = TCRain.from_tracks(tc_track, centroids=CENTR_TEST_BRB)
+        tc_haz = TCRain.from_tracks(tc_track, CENTR_TEST_BRB)
```
emanuel-schmid, line 69:
```suggestion
        tc_haz = TCRain.from_tracks(tc_track, centroids=CENTR_TEST_BRB)
```

climada_petals/hazard/test/test_tc_rainfield.py
```diff
@@ -117,12 +116,11 @@ def test_cross_antimeridian(self):
         # Two locations on the island Taveuni (Fiji), one west and one east of 180° longitude.
         # We list the second point twice, with different lon-normalization:
         cen = Centroids.from_lat_lon([-16.95, -16.8, -16.8], [179.9, 180.1, -179.9])
-        cen.set_dist_coast(precomputed=True)
 
         # Cyclone YASA (2020) passed directly over Fiji
         tr = TCTracks.from_ibtracs_netcdf(storm_id=["2020346S13168"])
 
-        inten = TCRain.from_tracks(tr, centroids=cen).intensity.toarray()[0, :]
+        inten = TCRain.from_tracks(tr, cen).intensity.toarray()[0, :]
```
emanuel-schmid, line 123:
```suggestion
        inten = TCRain.from_tracks(tr, centroids=cen).intensity.toarray()[0, :]
```

climada_petals/hazard/test/test_tc_rainfield.py
```diff
@@ -135,7 +133,7 @@ def test_cross_antimeridian(self):
     def test_from_file_pass(self):
         """Test from_tracks constructor with one input."""
         tc_track = TCTracks.from_processed_ibtracs_csv(TEST_TRACK_SHORT)
-        tc_haz = TCRain.from_tracks(tc_track, centroids=CENTR_TEST_BRB)
+        tc_haz = TCRain.from_tracks(tc_track, CENTR_TEST_BRB)
```
emanuel-schmid, line 136:
```suggestion
        tc_haz = TCRain.from_tracks(tc_track, centroids=CENTR_TEST_BRB)
```

climada_petals/hazard/test/test_tc_rainfield.py
```diff
@@ -159,7 +157,7 @@ def test_from_file_pass(self):
     def test_two_files_pass(self):
         """Test from_tracks constructor with two ibtracs."""
         tc_track = TCTracks.from_processed_ibtracs_csv([TEST_TRACK_SHORT, TEST_TRACK_SHORT])
-        tc_haz = TCRain.from_tracks(tc_track, centroids=CENTR_TEST_BRB)
+        tc_haz = TCRain.from_tracks(tc_track, CENTR_TEST_BRB)
```
emanuel-schmid, line 160:
```suggestion
        tc_haz = TCRain.from_tracks(tc_track, centroids=CENTR_TEST_BRB)
```

script/jenkins/branches/run_unittest.sh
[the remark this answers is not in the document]
peanutfun: Ah, sorry, I was very confused by this! If we can undo them, I would prefer that. We can discuss such changes in a separate PR, I think.

climada_petals/hazard/tc_rainfield.py
peanutfun: I see. `on_land` wasn't used before, as far as I can see, so we can leave as is.

climada_petals/hazard/river_flood.py
peanutfun: If we revert changes to this setup anyway, no worries :v:

climada_petals/hazard/test/test_wildfire.py
chahank: Actually, the method [Centroids.from_pnt_bounds](https://github.com/CLIMADA-project/climada_petals/blob/9f48d3fdae8d11d4a4c61212ff1a469d9cf1ef69/climada_petals/hazard/wildfire.py#L658) should produce exactly the same latitude and longitude values in both versions of the code. 

I think the point here is that we are now only using the centre of the raster pixels instead of the raster itself, there might be a half-pixel difference in the bounds. I am not sure to understand what this test is supposed to check... Why is the tolerance set to 0.005? Why should there be a tolerance? O really have to dig into this code to understand this.

emanuel-schmid: the tolerance is an artifact that I introduced simply to make the test pass, suspecting that the reason for failure may be something like what you mention - and that it does not matter.

chahank: aaaa got it! I will try to debug and find the reason.

chahank: So, the change has to do with the pixel definition. We have that 

`new_centroids.total_bounds[1] - 0.375/ONE_LAT_KM= old_centroids.total_bounds[1]`

This almost makes sense, as `0.375/ONE_LAT_KM` is the data raster resolution. But I would have expected a factor of one-half... However, the method `wf._firms_centroids_creation(firms, 0.375/ONE_LAT_KM, 1/2)` applies another factor to obtain the final raster resolution, here equal exactly to 1/2, which thus cancels the other one out.

chahank: Solution suggested in [116e2e8](https://github.com/CLIMADA-project/climada_petals/pull/122/commits/116e2e83d79c292735f65a74a7bc65c518c842aa)

emanuel-schmid: 🎉 excellent analysis! I'm still wondering though - shall we keep the test ast it is, do we change it or should we even got back and change the implementation?

emanuel-schmid: ok. 116e2e8 answers the question 😄

emanuel-schmid:
Unless I'm missing something we're ready for merging - right?

PR #142, opened 2024-09-03
Download and process Copernicus data to Hazard object

DahyannAraya:
Changes proposed in this PR:
- download

Note that the new module is placed in a new subdepository because further functionalities are planned to be added in the future.

This PR fixes #

### PR Author Checklist

### PR Reviewer Checklist


climada_petals/hazard/copernicus_forecast/handler.py
```diff
@@ -0,0 +1,560 @@
...
+2. CDS account and API key:
+   Register at https://cds-beta.climate.copernicus.eu
+
+3. CDS API configuration:
+   Create a .cdsapirc file in your home directory with your API key and URL.
+   For instructions, visit:
+   https://cds-beta.climate.copernicus.eu/how-to-api#install-the-cds-api-client
+"""
+
+import os
```
chahank, line 36: I would suggest to use Pathlib instead of os whenever possible.

climada_petals/hazard/copernicus_forecast/handler.py
```diff
@@ -0,0 +1,560 @@
...
+class ForecastHandler:
+    """
+    A class to handle downloading, processing, and calculating climate indices
+    and hazards based on seasonal forecast data from Copernicus Climate Data Store (CDS).
+    """
+
+    _FORMAT_GRIB = 'grib'
+    _FORMAT_NC = 'nc'
+    
+    def __init__(self, data_dir='.', url = None, key = None):
```
chahank, line 62: It might be more consistent to use the default data_dir defined in the configuration. See for instance the data_api.

climada_petals/hazard/copernicus_forecast/handler.py
```diff
@@ -0,0 +1,560 @@
...
+            try: 
+                north, west, south, east = area_selection
+                lat_margin = margin * (north - south)
+                lon_margin = margin * (east - west)
+                north += lat_margin
+                east += lon_margin
+                south -= lat_margin
+                west -= lon_margin
+                return [north, west, south, east]
+            except:
```
chahank, line 109: NEVER raise an exception and not handle it.

climada_petals/hazard/copernicus_forecast/handler.py
```diff
@@ -0,0 +1,560 @@
...
+                combined_bounds[1] = min(combined_bounds[1], min_lon - lon_margin)
+                combined_bounds[2] = min(combined_bounds[2], min_lat - lat_margin)
+                combined_bounds[3] = max(combined_bounds[3], max_lon + lon_margin)
+            
+            if combined_bounds == [-90, 180, 90, -180]:
+                return None
+            else:
+                return combined_bounds
+
+    def explain_index(self, tf_index):
```
chahank, line 136: Please use either `indicator` or `index` consistently in all the code.

climada_petals/hazard/copernicus_forecast/handler.py
```diff
@@ -0,0 +1,560 @@
...
+        system (str): The forecast system version.
+
+        Returns:
+        None
+        """
+        try:
+            # Ensure the directory exists before downloading
+            output_dir = os.path.dirname(filename)
+            if not os.path.exists(output_dir):
+                os.makedirs(output_dir, exist_ok=True)
```
chahank, line 210: Please use `Pathlib`

climada_petals/hazard/copernicus_forecast/handler.py
```diff
@@ -0,0 +1,560 @@
...
+    
+    def __init__(self, data_dir='.', url = None, key = None):
+        """
+        Initializes the ForecastHandler instance.
+
+        Parameters:
+        data_dir (str): Path to the directory where downloaded and processed data will be stored.
+            Defaults to the current directory ('.').
+        url (str): url to the CDS API. Defaults to None, in which case the url from /.cdsapirc
+            is used.
```
chahank, line 70: Why is there the need to give in the url? Is there more than one possible?

climada_petals/hazard/copernicus_forecast/handler.py
```diff
@@ -0,0 +1,560 @@
...
+import cdsapi
+
+from climada.hazard import Hazard
+from climada.util.coordinates import get_country_geometries
+import climada_petals.hazard.copernicus_forecast.indicator as indicator
+
+
+LOGGER = logging.getLogger(__name__)
+
+class ForecastHandler:
```
chahank, line 53: Why is it called ForecastHandler? Is this not simply a class to download any type of data from the Copernicus data store?

climada_petals/hazard/copernicus_forecast/handler.py
chahank, line 70: Please use NumpyDoc style for all docstrings.

climada_petals/hazard/copernicus_forecast/handler.py
```diff
@@ -0,0 +1,560 @@
...
+                return None
+            else:
+                return combined_bounds
+
+    def explain_index(self, tf_index):
+        """
+        Prints an explanation and input data for the selected index.
+
+        Parameters:
+        tf_index (str): The climate index identifier.
```
chahank, line 141: What does `tf` mean here?

climada_petals/hazard/copernicus_forecast/handler.py
```diff
@@ -0,0 +1,560 @@
...
+        try:
+            # Ensure the directory exists before downloading
+            output_dir = os.path.dirname(filename)
+            if not os.path.exists(output_dir):
+                os.makedirs(output_dir, exist_ok=True)
+            
+            # Initialize the CDS API client and attempt to download the file
+            c = cdsapi.Client(url=self.url, key=self.key)
+            c.retrieve(
+                'seasonal-original-single-levels',
```
chahank, line 215: Why is the date source hardcoded? What about non-seasonal forecast?

climada_petals/hazard/copernicus_forecast/indicator.py
```diff
@@ -0,0 +1,470 @@
...
+            "index_long_name": "Hot Days (Tmax > 30°C)",
+        },
+    }
+    return index_params.get(index)
+
+def calculate_relative_humidity_percent(t2k, tdk):
+    """
+    Calculates the relative humidity percentage from temperature and dewpoint temperature.
+
+    Parameters:
```
chahank, line 108: Please use numpy doc style docstrings.

climada_petals/hazard/copernicus_forecast/handler.py
DahyannAraya: As you suggested, I replaced os with pathlib for improved path handling and readability.

climada_petals/hazard/copernicus_forecast/indicator.py
DahyannAraya: Done!

climada_petals/hazard/copernicus_forecast/handler.py
DahyannAraya: The data source is hardcoded because, in the demo, we will only present seasonal forecast data, which is the main goal of the project and what we have committed to delivering. However, a few months ago, we discussed the possibility of making the part of the code responsible for the download flexible enough to access the Copernicus repository (Do not worry still in my mind). This is feasible, and I had discussed it in a meeting where I planned the module’s architecture with Lucas. But I will only implement this change once we have presented the demo, as this is what we are reporting this month in the semestral project report.

Afterwards, the idea is to gather all feedback and suggestions about the module during the CLIMADA days, including comments from DWD and C2SM, and then begin restructuring and optimizing the module. For now, unless something unexpected happens (which can always occur), we are on track with the schedule.

chahank: Understood. Thanks for the clarification.

climada_petals/hazard/copernicus_forecast/handler.py
DahyannAraya: Done

climada_petals/hazard/copernicus_forecast/handler.py
DahyannAraya: Currently, it’s called ForecastHandler because its main function is to download, process, and convert data into indices, which are then used to generate Hazard objects. However, this will evolve as the module is restructured. The naming of the scripts is still very flexible. Valentin and I considered options like 1. descargar.py, 2. forecast_index_downloader.py, and 3. indices.py (I know Lukas isn’t a fan of plurals). But this is entirely open for discussion, so if you have any suggestions, they would be more than welcome!

climada_petals/hazard/copernicus_forecast/handler.py
DahyannAraya: I have maintained the current structure for the Copernicus Data Store (CDS), which includes the option to use a configurable URL. However, CDS Beta, the new Copernicus website, has recently been launched last week, and many components are undergoing changes. The idea is to keep this setup until everything stabilizes and we have a clearer understanding of how to optimize downloads under the new circumstances since more things can change.

climada_petals/hazard/copernicus_forecast/handler.py
DahyannAraya: Done

climada_petals/hazard/copernicus_forecast/handler.py
chahank: My question was more about the word 'forecast'. Since this could also handle any type of weather data, not only forecasts.

climada_petals/hazard/copernicus_forecast/handler.py
chahank: OK, then I would put the default as the string instead of None. Also, it seems that this class attribute is never used later on.

climada_petals/hazard/copernicus_forecast/handler.py
DahyannAraya: For the demo, we are currently using only forecast data because the data architecture is specifically structured for seasonal forecasts, such as those from DWD, UK Met Office, Météo France, CMCC, NCEP, and JMA, which may differ from climate data. Seasonal forecast data is time-sensitive and focuses on predicting short- to medium-term trends (weeks to a maximum of 7 months) based on initialized models. It takes into account specific initial conditions, like the forecast month and hourly resolution, to build indices such as heatwaves or temperature extremes (e.g., TX30, TR).

If we want to calculate indices for other datasets, we need to first study the architecture of that data and then test it. For now, only seasonal forecast data can be handled effectively.

ValentinGebhart: The current class only handles forecast data from a specific seasonal forecast. We plan to add a more general class (e.g. "CopernicusDownloader") that simply downloads any copernicus data set (and saves it to netcdf eg). The current forecast handler then could then use this download class and processes the data. 
A general "handler" that calculates indices for both forecast and other data types might be more difficult because the downloaded data set probably have very different structures, for instance the member dimension of forecasts. Does this make sense?

climada_petals/hazard/copernicus_forecast/handler.py
ValentinGebhart: We put it to None because when handing "None" to the client.retreive function of the cdsapi module, the client will use the url and key that are saved in a separate file in the user's home directory (see instructions). If I understood correctly, this file is necessary for the downloader to work, so the user must add it anyways.

climada_petals/hazard/copernicus_forecast/handler.py
chahank: Yes, thanks for clarifying.

climada_petals/hazard/copernicus_forecast/handler.py
chahank: Ok, interesting. Then the docstring is incorrect. Overall though, I think this could be handled more smoothly.

ValentinGebhart: Also to add, Dahyann said that DWD had specific requests on what the user should input when they want to download something (they wanted that the users have the cdsapi config file and ALSO have to put in url and key by hand). Then, as Dahyann said, the API is currently changing and we don't know the final form. 
So we agreed that we leave it as is for now and when we know all boundary conditions, we could optimize further. I agree with you that it could be handled better

chahank: I understand that, and it is good to listen to DWD. At the same time in the end, this is a CLIMADA module, not a DWD module. Thus, we should still keep a sensible coding setting whenever possible.

chahank: Here it is a question of design, and not really of optimization. Thus I would caution against using the argument: "we will just fix it later".

ValentinGebhart: ok agreed, we will have a chat about better possibilities.

climada_petals/hazard/copernicus_forecast/handler.py
DahyannAraya: "tf_index" was taken from the original code discussed at C2SM. To clarify, I have updated the name to "index_metric," allowing selection from TX30, TR, HIS, HIA, etc.

climada_petals/hazard/copernicus_forecast/handler.py
DahyannAraya: I removed the URL parameter to avoid confusion. The default .cdsapirc configuration is sufficient for standard use since there's "usually" only one .cdsapirc URL.

climada_petals/hazard/copernicus_forecast/handler.py
DahyannAraya: Done

climada_petals/hazard/copernicus_forecast/handler.py
DahyannAraya: I simplified the configuration by removing the URL and key parameters, defaulting to .cdsapirc for consistency

climada_petals/hazard/copernicus_forecast/handler.py
DahyannAraya: I keep indicator

climada_petals/hazard/copernicus_forecast/handler.py
DahyannAraya: Thanks for the comments! The name fits right now since we handle only forecast data, but we’re open to changing it as the module evolves. As was mentioned, we will discuss adding a general downloader class to support other data types, and if that happens, I will adjust the naming accordingly.

emanuel-schmid (commented):
Many thanks! Already looks quite decent to me. 😄 

**General Suggestions**

- let the handler organize all paths, deal with `overwrite`, keep track of parameters
- move the year-month loops out of download, process, calc_index and to_hazard and up one or two levels
- whenever a function creates a (list of) file(s), it should return them
- use files as input arguments

- use dataclasses and Enums instead of dict constants
- use `climada_petals.CONFIG` for storage path

climada_petals/hazard/copernicus_forecast/create_seasonal_forecast_hazard.py
```diff
@@ -0,0 +1,993 @@
...
+        hazard_obj.plot_intensity(1, smooth=False)
+        converted_dates = [dt.datetime.fromordinal(date) for date in hazard_obj.date]
+        plt.figtext(
+            0.5,
+            0.23,
+            f"Date: {converted_dates[0].strftime('%Y-%m-%d')}",
+            ha="center",
+            fontsize=16,
+        )
+        plt.tight_layout()
```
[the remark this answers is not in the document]
DahyannAraya: The display is no longer available in the hazard function calculation.

climada_petals/hazard/copernicus_forecast/seasonal_statistics.py
```diff
@@ -0,0 +1,650 @@
...
+        "short_name": "v10",
+        "full_name": "10m_v_component_of_wind",
+    },
+    "10m_wind_gust_since_previous_post_processing": {
+        "unit": "m/s",
+        "standard_name": "wind_gust",
+        "short_name": "wind_gust10m",
+        "full_name": "10m_wind_gust_since_previous_post_processing",
+    },
+}
```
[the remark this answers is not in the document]
DahyannAraya: I created a new index_definitions.py file with a @dataclass to structure index metadata. As you suggested, the IndexSpec dataclass encapsulates properties such as units, variables, names, and explanations for each index. Additionally, the items are organized under an IndexSpecEnum.

climada_petals/hazard/copernicus_forecast/create_seasonal_forecast_hazard.py
```diff
@@ -0,0 +1,993 @@
...
+            if m > 12:
+                adjusted_year += 1
+                adjusted_month = m - 12
+
+            num_days_month = calendar.monthrange(adjusted_year, adjusted_month)[1]
+            timesteps = num_days_month * 24
+            total_timesteps += timesteps
+
+        max_lead = total_timesteps + 6
+        return 0, max_lead
```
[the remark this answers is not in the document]
DahyannAraya: Thank you for the suggestions! I opted for the approach where the entire lead time calculation is encapsulated in a single method that returns a list of lead times ([0, 6, ... max_lead]). This aligns with your suggestion to make the method more general by avoiding the tuple (0, something) and instead iteratively generating lead times in 6-hour intervals, as you said. This approach avoids hardcoding the minimum lead, also offer flexibility

climada_petals/hazard/copernicus_forecast/heat_index.py
```diff
@@ -0,0 +1,420 @@
...
+    # Initialize array to store heatwave days, filled with 0s
+    hw_days = np.zeros(len(temperatures), dtype=int)
+
+    # Create a binary array: 1 for temperatures > threshold, 0 for ≤ threshold
+    binary_array = np.where(temperatures >= threshold, 1, 0)
+
+    events = []
+    prev_continous_ones = 0
+
+    for i in range(len(binary_array)):
```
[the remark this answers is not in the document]
DahyannAraya: I’ve updated the code to use enumerate for iterating through binary_array, replacing range(len(...)). This makes the loop cleaner and more readable, as recommended. Thanks

ValentinGebhart:
This module has been moved to https://github.com/DahyannAraya/copernicus-seasonal-forecast-tools.

PR #150, opened 2024-11-27
Added function to download data from Copernicus data stores

ValentinGebhart:
This is one part of the `copernicus_interface` module that is currently under development (see PR #142).

Changes proposed in this PR:
- Added function `download_data` to `climada_petals.hazard.copernicus_interface` module that accesses a Copernicus data store and downloads a dataset with user-given parameter ranges.

Notes and questions:
- We have thought about what unit tests to write but, expect for some type checking of the input parameters, we could not think of a useful unit test. We do not want the unit test to actually access a datastore. Any suggestions welcome!

### PR Author Checklist

### PR Reviewer Checklist

- [x] [CLIMADA Reviewer Checklist](https://climada-python.readthedocs.io/en/latest/guide/Guide_Reviewer_Checklist.html) passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Added:
 ### Added
 
 - Included `boario` in the supplychain module [#81](https://github.com/CLIMADA-project/climada_petals/pull/81/)
+- Added a Copernicus download function `downloader.py` as part of the `climada.hazard.copernicus_interface` module under construction [#150](https://github.com/CLIMADA-project/climada_petals/pull/150/)
 
 ### Changed
 
--- a/climada_petals/conf/climada.conf
+++ b/climada_petals/conf/climada.conf
@@ -54,6 +54,10 @@
                     "passwd": "essential"
                 }
             }
+        },
+        "copernicus": {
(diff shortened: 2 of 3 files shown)


luseverin (commented):
Nice work! I tried downloading different variables from different datasets and it worked well. Besides what I have commented in the code directly, I have a few more general comments:
- I think there are still some improvements that could be made in terms of user-friendliness e.g. saying where the names and descriptions of the variables necessary for the call to the api can be found. I made a few comments on that  in the code.
- I did a few tests to test your handling of errors but did not try everything. The errors I got from the data provider directly were still sometimes a bit cryptic, but I think that in addition to the warnings and errors that you log I always had enough information to figure what was going wrong.
- I think some errors are still passed silently; for instance I can enter anything for the `data_format` variable but I always get something. But probably this is more an issue for the data provider than for you.
- A suggestion I have made earlier would be to log a warning if the queue is too long, but I am not sure if the information of  the place in the queue for the user's query can be retrieved easily.
- Regarding unit tests, I do not have other suggestions besides testing the type of the inputs.
- There are some unit tests failing but I think they are related to anothe module (something with the drought module )

PR #153, opened 2024-12-13
Feature/add missing jrc flood impact functions

aleeciu:
This PR is a follow up of https://github.com/CLIMADA-project/climada_petals/pull/111 where I fix the failing integration test. 

Two things to note about the fix:
- the impf_code changed from 3 to 31
- the function itselft slightly changed to ensure consistency across regions. In particular, a point was added with mdd = 0 and intensity = 0.05. This changes the tested damage value which was also updated.


aleeciu:
I seem to be getting test errors about wind calculation. I don't know where this comes from.

emanuel-schmid:
@aleeciu : those wind calculation errors came from a bad commit 😜 
Now the only failing checks are about code coverage: the total number of tested lines has apparently decreased.
I've checked: the only lines related to this PR that are not covered are the ones in the deprecated methods, e.g., `set_RF_Impf_SouthAmerica`. Bottom line: all checks were _essentially_ successful.

aleeciu:
@emanuel-schmid so would you think I can merge? :-)

emanuel-schmid:
Er.. actually it should be reviewed first. I just checked whether it is even ready for review. 😁

emanuel-schmid:
@aleeciu Many thanks for fixing the tests and the explanations about why this changed!
I'll merge this.

Issue #164, opened 2025-03-13
download_glofas_discharge fails with "ValueError: did not find a match in any of xarray's currently installed IO backends"


luseverin:
I am trying to download discharge forecasts using `download_glofas_discharge` directly.
```
from climada_petals.hazard.rf_glofas.transform_ops import download_glofas_discharge

countries : ["KEN"],
forecast_date = "2024-03-20"
lead_time_days = 5
preproc = lambda x: x.max(dim="step").mean(dim="number")
leadtime_hour = list(
            map(str, (np.arange(1, query_forecast["lead_time_days"] + 1, dtype=np.int_) * 24).flat)
)
forecast = download_glofas_discharge(
    product="forecast",
    date_from=,
    date_to=None,
    countries=countries,
    preprocess=preproc,
    leadtime_hour=leadtime_hour,
    system_version="version_3_1",
)
```
However, it fails with a ValueError:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[14], [line 8](vscode-notebook-cell:?execution_count=14&line=8)
      [2](vscode-notebook-cell:?execution_count=14&line=2) from climada_petals.hazard.rf_glofas.transform_ops import download_glofas_discharge
      [5](vscode-notebook-cell:?execution_count=14&line=5) leadtime_hour = list(
      [6](vscode-notebook-cell:?execution_count=14&line=6)             map(str, (np.arange(1, query_forecast["lead_time_days"] + 1, dtype=np.int_) * 24).flat)
      [7](vscode-notebook-cell:?execution_count=14&line=7) )
----> [8](vscode-notebook-cell:?execution_count=14&line=8) forecast = download_glofas_discharge(
      [9](vscode-notebook-cell:?execution_count=14&line=9)     product="forecast",
     [10](vscode-notebook-cell:?execution_count=14&line=10)     date_from=pd.Timestamp(query_forecast["forecast_date"].min()).date().isoformat(),
     [11](vscode-notebook-cell:?execution_count=14&line=11)     date_to=pd.Timestamp(query_forecast["forecast_date"].max()).date().isoformat(),
     [12](vscode-notebook-cell:?execution_count=14&line=12)     countries=query_forecast["countries"],
     [13](vscode-notebook-cell:?execution_count=14&line=13)     preprocess=query_forecast["preproc"],
     [14](vscode-notebook-cell:?execution_count=14&line=14)     leadtime_hour=leadtime_hour,
[...]

Issue #141, opened 2024-09-03
Change hazard.frequency automatically to hazard.frequency/hazard.size during hazard computation in rf_glofas 


elianekobler:
The default hazard.frequency value of 1 leads to excessive impact numbers when plotting impacts from 50 ensemble members (GloFAS Forecast). The _build_exp() function is used during plotting, which takes the eai_exp value at each centroid. Per centroid all 50 impact values (50 events) are summed up each with a frequency of 1. This leads to unexpected high impact values per centroid.
In the tutorial it is described like this:
_"The averaging function takes the `frequency` parameter of the hazard into account.
By default, all hazard events have a frequency of 1.
This may result in unexpected values for average impacts.
We therefore divide the frequency by the number of events before plotting."_

The issue occurs when you follow the climada_hazard_glofas_rf tutorial but forget to add this line before the impact calculation:
`hazard.frequency = hazard.frequency / hazard.size`

The simplified calculation steps are as follows:
```
rf = RiverFloodInundation()
discharge = rf.download_forecast()
ds_flood = rf.compute(discharge=discharge)
save_file(ds_flood)
with xr.open_dataset(f"flood-2024-08-04.nc", chunks="auto") as ds:
    hazard = hazard_series_from_dataset(ds, "flood_depth", "number")
impact = ImpactCalc(exposure, impf_set_affected, hazard).impact()
impact.plot_hexbin_eai_exposure()
```
This is how my plots looked like: (the mean exposed population is calculated from impact.at_event values and therefore not affected by the frequency)
<img width="813" alt="Bildschirmfoto 2024-09-03 um 10 46 14" src="https://github.com/user-attachments/assets/3c7def23-4f16-4bfb-a0a5-70d7642e50aa">

Would it be a better solution to divide the hazard.frequency by the hazard.size automatically during the hazard_series_from_dataset() computation? 
Another solution would be to return a warning to the user as they may want to adjust the hazard.frequency.

peanutfun:
It sounds reasonable to set the frequency to 1/size automatically!

PR #167, opened 2025-03-21
Update RiverFlood module to GloFAS v4

### LINKED ISSUE #139: Check if CDS downloader of GloFAS river flood module is compatible with new CDS Engine API
See https://confluence.ecmwf.int/display/CKB/Please+read%3A+CDS+and+ADS+migrating+to+new+infrastructure%3A+Common+Data+Store+%28CDS%29+Engine

* Upgrade to cdsapi v0.7 compatible?
* Do we need to change download requests?

peanutfun:
Changes proposed in this PR:
- Update to GloFAS v4
- Change handling of flood hazard maps: Download and open individual tiles instead of merging during setup
- Add option to split download requests based on time parameter within the data
- Update docs and tests

This PR fixes #139 #164 #141

### PR Author Checklist

### PR Reviewer Checklist


luseverin:
Thanks a lot @peanutfun for taking the time to update the model! I tried the updated version and it seems to be working quite well so far. I just wanted to report one issue here:
I get a `HTTPError`:
```
HTTPError: 400 Client Error: Bad Request for url: https://ewds.climate.copernicus.eu/api/retrieve/v1/processes/cems-glofas-forecast/execution
invalid request
No valid dates from year=[2024] month=[6] day=[31]
```

Apparently, the issue comes from the way you format the dates in `cds_glofas_downloader.glofas_request`:
```
requests = [
     {**default_request, **dict(zip(split_request_map.keys(), bits))}
     for bits in it.product(*split_request_map.values())
]
```
As it seems to me, doing this you assume that each month has the same number of days, creating the error. A fix could be to catch this error specifically and pass when the requested day does not exist. Otherwise, maybe limit the number of days for unpair months, what do you think?

peanutfun:
@luseverin good catch! I think the problem is in `datetime_index_to_request`, where days, months and years are considered separately. Can you tell me exactly which call to `download_glofas_discharge` or `RiverFloodInundation.download_forecast` lead to the error?

luseverin:
Thanks for looking into it! I think the following call resulted in the error:
```
start_date = f"{2024}-{3}-{20}"
end_date = f"{2024}-{6}-{10}"

forecast_dates = pd.date_range(start_date, end_date)
lead_time_days = 2
leadtime_hour = list(
    map(str, (np.arange(1, lead_time_days + 1, dtype=np.int_) * 24).flat)
)
#download forecast
forecast_ds = download_glofas_discharge(
    product="forecast",
    dates=forecast_dates,
    countries=["Kenya"],
    preprocess=lambda x:x.max(dim="forecast_period").mean(dim="number", keepdims=True),
    leadtime_hour=leadtime_hour,
    system_version="operational",
    num_proc=8
)
```
Let me know if you manage to replicate this error.

elianekobler:
Thanky you @peanutfun for the updated glofas rf model!

After running setup_all() I only got the FLOPROS dataset and the gumbel-fit.nc . However, it is still mentioned in the docstring that setup_all() would download and merge JRC river flood hazard maps. Is the setup_flood_hazard_maps() function excluded from the setup_all() function on purpose?

peanutfun:
@elianekobler, yes, this is on purpose. The docs are outdated. The merging of the flood maps is not done anymore because the merged dataset is too large. The individual flood map tiles are downloaded by `RiverFloodInundation` when required.

luseverin:
@elianekobler and I have been recurently running into connection error when downloading discharge from the CDS api:
```
WARNING Recovering from connection error [('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))], attemps 0 of 500
2022-03-18 22:05:30,589 WARNING Retrying in 120 seconds
```
In some cases, it sometimes reconnected after some time but in other cases it was seemingly running forever..
For, me reducing the number of processes (`num_proc=1`) seems to have make things a bit more stable, but it still sometimes occurs. Not sure if this related to unstable connection (however, I've never had this issue with glofas3) or to something else e.g. an excessive number of queries to the server. @peanutfun have you encountered similar issues as well?

peanutfun:
@luseverin I frequently encountered timeout issues when downloading the historical data on Euler. It seemed to me like a single process would stall the download when running in parallel. However, I could not narrow the problem down. It might be related to the CDSAPI (after the switch to the new Emergency Data Store), or with running the requests in parallel.

luseverin:
> Thanks a lot @peanutfun for taking the time to update the model! I tried the updated version and it seems to be working quite well so far. I just wanted to report one issue here: I get a `HTTPError`:
> 
> ```
> HTTPError: 400 Client Error: Bad Request for url: https://ewds.climate.copernicus.eu/api/retrieve/v1/processes/cems-glofas-forecast/execution
> invalid request
> No valid dates from year=[2024] month=[6] day=[31]
> ```

I added a fix to get rid of this error. However, there is still the issue that the data returned to the user does not match the queried dates (i.e. if you ask for flood forecasts for the period 2024-01-31 to 2024-02-01 you will obtain forecasts for the 2024-01-31 , 2024-02-01  **and** 2024-01-01). This might be problematic especially if one does not check what exactly comes out of the request and directly compute statistic over the entire dataset.

peanutfun:
@luseverin regarding the date issue, I did some tests and found out the following:
1. It is impossible to download data for particular days with a months switch in one request: Requesting for 2024-01-31 and 2024-02-01 yields the request 
   ```python
   {"month": ["01", "02"],
    "day": ["31", "01"]}
   ```
   which actually downloads 2024-01-01, 2024-01-31, 2024-02-01, 2024-02-31. Therefore, one always has to split the requests in such a case
2. We do not have to limit the call to `unique` days/months/years because the CDSAPI does so automatically. The following downloads 2024-01-01 amd 2024-01-02:
   ```python
    {"year": ["2024", "2024", "2024"],
     "month": ["01", "01"],
     "day": ["01", "02"]}
   ```

Consequently, we can remove the call to `unique` day/month/year values and make the users aware of the API shortcomings (point 1). With `split_request_keys` it should be possible to find a combination of requests that downloads exactly what you want/need

luseverin:
Hi @peanutfun, I just had a look and indeed the splitting by day fixes the issue of downloading unwanted dates, well done!

peanutfun:
@luseverin I cannot replicate this error. I just downloaded [the file](https://www.research-collection.ethz.ch/entities/researchdata/be75442a-93d4-4b4e-864c-82bd07c575d7) and I can open it without problems. Can you check if the file is complete? It has a size of 106 MB on my machine. Maybe there is an issue with the download routine for the file.

Please note that this part of the tutorial likely won't work. The forecast from 2021-07-10 was conducted with GloFAS v3, so you will get an error later on when you try to compute the return period. For 2021-07-10, you will have to use reforecast data that was computed with GloFAS v4. I discussed this with @elianekobler yesterday.

peanutfun:
The research collection has a new-looking frontend. The download link probably changed. I'll have a look

peanutfun:
@luseverin The link has changed indeed. I just updated the branch

luseverin:
> @luseverin The link has changed indeed. I just updated the branch

Thanks! It works now.

peanutfun:
We need to handle the case where the discharge dataset only contains the spatial dimensions. The data pipeline currently expects at least one other dimension.

peanutfun:
Todo: Support 2D discharge datasets

climada_petals/hazard/rf_glofas/flood_maps.py
```diff
@@ -67,7 +67,7 @@ def tile_url(return_period: int, tile_id: int, name: str) -> str:
     :py:func:`open_flood_maps_extents`
         The function that loads the tile extents dataset.
     """
-    return JRC_FLOOD_HAZARD_MAP_URL.format(rp=return_period, tile_id=tile_id, name=name)
+    return JRC_FLOOD_HAZARD_MAP_URL.format(rp=return_period, tile_id=tile_id+1, name=name)
```
peanutfun: @luseverin Are you sure about this? The file contents look like this
```
In [2]: gpd.read_file("https://jeodpp.jrc.ec.europa.eu/ftp/jrc-opendata/CEMS-GLOFAS/flood_hazard
      ⋮ /tile_extents.geojson")
Out[2]: 
      id      name                                           geometry
0      1  N70_W180  POLYGON ((-180.00042 70.00207, -170.00125 70.0...
1      2  N80_W170  POLYGON ((-170.00125 80.00123, -160.00208 80.0...
2      3  N70_W170  POLYGON ((-170.00125 70.00207, -160.00208 70.0...
3      4  N60_W170  POLYGON ((-170.00125 60.0029, -160.00208 60.00...
4      5  N80_W160  POLYGON ((-160.00208 80.00123, -150.00292 80.0...
..   ...       ...                                                ...
```
And the file directory looks like this:
Image

Why would you need to increment the tile ID?

peanutfun: I think this was mainly concidence because a new tile was inserted in the beginning. The fix is to update the tests, because the original tiles received new IDs, not to update the link (which will cause the wrong files to be downloaded)

luseverin: Hi @peanutfun thanks for looking into it. Sorry if I applied the wrong fix and thanks for correcting my mistake. Is there anything else which needs to be done? Regarding this specific issue or the update in general.

peanutfun: I think we are good!

peanutfun:
@emanuel-schmid This seems to be working well for @luseverin and @elianekobler, could you do a review so we can merge this soon and fix the failing compatibility tests from core?

emanuel-schmid:
I can already tell so much: the changes in the env_climada.yaml file are not compatible with Windows OS. 😟 
So far the gfortran dependency has been excluded from the environment file for this very reason. Perhaps it is even possible to install gfortran within a conda environment for windows, but certainly not together with the other given dependencies.
Our approach how to deal with was opportunistic. Users who want to use the module need to install gfortran separately, by themselves.
Afaik there is no way to specify optional dependencies in a conda env file, but we could have several os - specific env files.
Until now we didn't want to do that.

Bottom line: we have abandoned fortran for windows and we write the env-file so that it fits all OS.

climada_petals/hazard/rf_glofas/setup.py
```diff
@@ -211,28 +201,32 @@ def download_gumbel_fit(output_dir=DEFAULT_DATA_DIR):
 
 def setup_all(
     output_dir: Union[str, Path] = DEFAULT_DATA_DIR,
+    compute_gumbel_fit: bool = False,
+    **setup_gumbel_fit_kwargs,
```
emanuel-schmid: name cosmetics: what about `compute_gumbel_fit_kwargs`

climada_petals/hazard/rf_glofas/setup.py
```diff
@@ -211,28 +201,32 @@ def download_gumbel_fit(output_dir=DEFAULT_DATA_DIR):
...
     output_dir.mkdir(exist_ok=True)
 
-    setup_flood_hazard_maps(
-        flood_maps_dir=DEFAULT_DATA_DIR / "flood-maps", output_dir=output_dir
-    )
     download_flopros_database(output_dir)
-    download_gumbel_fit(output_dir)
+
+    if compute_gumbel_fit:
+        setup_gumbel_fit(output_dir, **setup_gumbel_fit_kwargs)
```
emanuel-schmid: name cosmetics: what about `compute_gumbel_fit(...)`  and change the parameter name?

climada_petals/hazard/rf_glofas/setup.py
```diff
@@ -170,37 +97,100 @@ def setup_gumbel_fit(
...
 
 
 def download_gumbel_fit(output_dir=DEFAULT_DATA_DIR):
     """Download the pre-computed Gumbel parameters from the ETH research collection.
 
-    Download dataset of https://doi.org/10.3929/ethz-b-000641667
+    Download dataset of https://doi.org/10.3929/ethz-b-000726304
     """
     LOGGER.debug("Downloading Gumbel fit parameters")
     response = requests.get(GUMBEL_FIT_DATA, stream=True)
```
emanuel-schmid:
```suggestion
    response = requests.get(GUMBEL_FIT_DATA, stream=True)
    if response.status_code != 200:
        raise RuntimeError(
            f"Failed to download Gumbel fit parameters from {GUMBEL_FIT_DATA}. Status code: {response.status_code}"
        )
```

climada_petals/hazard/rf_glofas/setup.py
```diff
@@ -170,37 +97,100 @@ def setup_gumbel_fit(
...
-        fit.to_netcdf(output_dir / "gumbel-fit.nc", engine="netcdf4")
 
 
 def download_gumbel_fit(output_dir=DEFAULT_DATA_DIR):
     """Download the pre-computed Gumbel parameters from the ETH research collection.
 
-    Download dataset of https://doi.org/10.3929/ethz-b-000641667
+    Download dataset of https://doi.org/10.3929/ethz-b-000726304
     """
     LOGGER.debug("Downloading Gumbel fit parameters")
```
emanuel-schmid, line 200: If the file is already there, the download could be skipped.
Especially whith the research collection re-organizing this could come in handy.
We'd need a way to check for correctness of the found file.

climada_petals/hazard/rf_glofas/transform_ops.py
```diff
@@ -317,25 +325,32 @@ def download_glofas_discharge(
...
         product=product,
-        date_from=date_from,
-        date_to=date_to,
         num_proc=num_proc,
         output_dir=download_path,
         request_kw=request_kwargs,
+        requests=requests,
     )
 
+    if open_mfdataset_kw is False:
```
emanuel-schmid: From the programming perspective this is imo a design flaw.
Any function should always return exactly one return type.
In this case likely files at the cost of having a follow up function that deals with the transformation from files to datasets.

doc/tutorial/climada_hazard_glofas_rf.ipynb
```diff
@@ -573,149 +555,181 @@
...
-    "    preprocess=lambda x: x.max(dim=\"step\"),\n",
-    "    system_version=\"operational\",  # Version mislabeled\n",
-    ")\n",
+    "# dis = rf.download_reanalysis(\n",
+    "#     countries=countries,\n",
+    "#     year=2021,\n",
+    "#     preprocess=lambda x: x.sel(time=slice(\"2024-07-10\", \"2024-07-20\")).max(dim=\"time\"),\n",
+    "# )\n",
+    "with xr.open_dataset(\n",
+    "    \"/Users/ldr.riedel/climada/data/cds-download/250627-105456-aa069b92e8690163ba2dc6d2213cb4d9.grib\",\n",
```
emanuel-schmid:
```suggestion
    "    f\"{{SYSTEM_DIR}}/cds-download/250627-105456-aa069b92e8690163ba2dc6d2213cb4d9.grib\",\n",
```

doc/tutorial/climada_hazard_glofas_rf.ipynb
```diff
@@ -51,34 +41,9 @@
   },
   {
```
emanuel-schmid, line 50: `setup_all()` succeeds but only downloads an error page as Gumbel_fits. The request persitently fails with HTTPError 500, "The server is temporarily unable to service your request due to maintenance downtime or capacity problems. Please try again later."
(see suggestion in setup.py)

`setup_all(compute_gumbel_fit=True)` on the other hand fails with HTTPError 404, "Client Error: Not Found for url: https://cds.climate.copernicus.eu/api/retrieve/v1/processes/cems-glofas-historical
process not found \
dataset cems-glofas-historical not found"

emanuel-schmid (changes requested):
I can't get the Tutorial to run. It fails at cell 2.

- if I run it as is, the .grib file is missing
- if I uncomment the lines ahead it fails with 'dataset cems-glofas-historical not found'

i.g. commented lines are fine with me, but I would like to get some infos about why they are there and what I can do with them.

I submit the review for now, because I think it's going to be a lot easier to continue when the tutorial issues are solved or clarified. 
to be continued ...

peanutfun:
> I can already tell so much: the changes in the env_climada.yaml file are not compatible with Windows OS. 😟

@emanuel-schmid The issue is that the conda package of esmf (a dependency of xesmf) currently does not support Windows because the devs cannot figure out the build recipe. See https://github.com/conda-forge/esmf-feedstock/issues/115

How did you manage to get it installed?

climada_petals/hazard/rf_glofas/setup.py
peanutfun: We usually only check if a file exists, not if it is correct, which is error-prone. We could introduce [pooch](https://www.fatiando.org/pooch/latest/about.html) as a dependency, which automatically verifies the checksum. Would you like to do that?

peanutfun: Clarification: Would **you** like **me** to do that? 😉

doc/tutorial/climada_hazard_glofas_rf.ipynb
peanutfun: The second error is probably because of an outdated config. You will need to update the `.cdsapirc`. The URL has changed. See https://climada-petals--167.org.readthedocs.build/en/167/glofas_rf.html#copernicus-data-store-api-access

peanutfun: The first error is an issue with the research collection. I uploaded the file to Zenodo now.

doc/tutorial/climada_hazard_glofas_rf.ipynb
peanutfun: Fixed

climada_petals/hazard/rf_glofas/transform_ops.py
peanutfun: I split the function into one downloading the data and one opening the data.

climada_petals/hazard/rf_glofas/setup.py
peanutfun: I added pooch as dependency for Petals

climada_petals/hazard/rf_glofas/setup.py
peanutfun: Done

climada_petals/hazard/rf_glofas/setup.py
peanutfun: Done

peanutfun:
Coarsen return period with max/mean aggregation automatically to account for high GloFAS v4 resolution?

elianekobler:
I run the setup_all() function without any error messages. However, the download of the gumbel-fit.nc file was restricted by ETH research collection. They provided the following message: "Access Restricted. Due to a high volume of automated traffic (scraping), access to the ETH Zurich Research Collection is temporarily restricted from your location / your provider. This measure helps ensure system availability for our users. If you have been blocked accidentially, please contact us at research-collection@library.ethz.ch."
Manual download works and they are trying to figure out the problem.

I also realised that the doi in the setup.py is pointing to the old gumbel-fit file from 2023. 
old: https://doi.org/10.3929/ethz-b-000641667 
new:  https://doi.org/10.3929/ethz-b-000726304

peanutfun:
@elianekobler Please update to the latest version. I uploaded a copy of the fit dataset to Zenodo, and changed the download URL accordingly in bbcc64cf895b5d206b81e2be71bdf74fec4ca3f3

peanutfun:
[interpolate_na](https://corteva.github.io/rioxarray/stable/rioxarray.html#rioxarray.raster_array.RasterArray.interpolate_na) only works for 2D and 3D arrays. How to extend to any dim?

peanutfun:
Use apply_func like here: https://github.com/MeteoSwiss/meteodata-lab/blob/c9be13a79adbdda8ed7433dbffb40b34156d1465/src/meteodatalab/operators/regrid.py#L308

PR #186, opened 2025-11-03
Create CAT bond module

KaiOBerg:
Changes proposed in this PR:

- add ibrd_bonds.csvto data/cat_bonds. Contains metrics of CAT bonds issued by the World Bank. Used to calculate premiums for CAT bonds in the cat_bond module. 
- add full CAT bond module, capable of siulating catastrophe bonds using different hazard data, exposure data and impact functions.

### PR Author Checklist

### PR Reviewer Checklist


climada_petals/engine/cat_bonds/subarea_calculations.py
```diff
@@ -0,0 +1,436 @@
...
+        subareas = range(len(haz_int[hazard_type].columns) - 2)
+        subarea_specific_results = {}
+
+        results = {}
+        for subarea in subareas:
+
+            damages = imp_subarea_evt_flt.iloc[:, subarea]
+
+            # Perform optimization for each subarea
+            result = minimize(
```
KaiOBerg, line 322: @spjuhel: we discussed that I should remove the initial guess in the optimization of the payout functions per subarea. However, it is a mandatory input to the minimize function. Do you have any suggestions?

spjuhel: Isn't there a default guess that would always be a valid first guess and do not need to store ?

climada_petals/engine/cat_bonds/mlt_bond_simulation.py
```diff
@@ -0,0 +1,244 @@
...
+                cties = events['country_code'].to_numpy()
+                pay = events['pay'].to_numpy()
+                dam = events['damage'].to_numpy()
+
+                sum_payouts = np.zeros(len(events))  
+                sum_damages = np.zeros(len(events)) 
+                for o in range(len(events)):
+                    payout = pay[o]
+                    cty = cties[o]
+                    damage = dam[o]
```
spjuhel:
```suggestion
                for payout, country, damage in zip(pay, countries, damages):
```

climada_petals/engine/cat_bonds/test/test_sng_bond.py
```diff
@@ -0,0 +1,129 @@
...
+    rel_losses, df_month, tot_pay, tot_dam = sim.init_bond_loss([df])
+
+    # principal exhausted at payout 2: payout[2] becomes (100 - 80) = 20
+    assert np.allclose(rel_losses.values, [100/100])  
+    assert np.allclose(tot_pay, 100)
+    assert tot_dam == 40 + 60 + 70
+
+    # Monthly losses divided by principal
+    assert df_month["losses"].iloc[0] == [0.30, 0.50, 0.20]
+    assert df_month["months"].iloc[0] == [1, 2, 3]
```
chahank: You can compare pandas dataframes with `pandas.testing.assert_frame_equal`.

climada_petals/engine/cat_bonds/test/test_premium_class.py
```diff
@@ -0,0 +1,53 @@
...
+import logging
+from climada_petals.engine.cat_bonds.premium_class import PremiumCalculations  
+
+logging.basicConfig(
+     format="{asctime} - {levelname} - {message}",
+     style="{",
+     datefmt="%Y-%m-%d %H:%M",
+     level=logging.INFO,
+ )
+LOGGER = logging.getLogger(__name__)
```
chahank: I am not sure where this comes frome. Can you please use the same testing environment as is used otherwise in CLIMADA?

KaiOBerg: updated LOGGER to logger saved in climada_petals//util/config.py

climada_petals/engine/cat_bonds/subarea_calculations.py
KaiOBerg: initial guess no depends on the distribution of given hazard intensity

climada_petals/engine/cat_bonds/pooling_functions.py
```diff
@@ -0,0 +1,297 @@
...
+    if x.sum() == 0:
+        POOL_CONC = 1.
+    else:
+        ES_tot = np.nansum(MES)
+        POOL_CONC = ES_tot / np.nansum(ES_cntry)
+
+    return np.round(POOL_CONC, 2)
+
+'''Pool optimization problem using fixed number of pools'''
+class PoolOptimizationFixedNumber(ElementwiseProblem):
```
KaiOBerg: @spjuhel I struggle to rewrite this class so that it doesn't inherit from pymoos ElementwiseProblem. I use this class to create the problem in the above written function process_n_pools. The problem is then used in pymoos minimize function which expects a pymoo project and a pymoo algorithm. Do you have a suggestion?
Thank you for your help!

climada_petals/engine/cat_bonds/pooling_functions.py
```diff
@@ -0,0 +1,297 @@
...
+        super().__init__(
+            n_var=self.data_arr.shape[1],
+            n_obj=1,  
+            n_constr = 1,
+            xl=0,                  
+            xu=self.N - 1,
+            type_var=int,
+            vars=[Integer((0, self.n_countries - 1)) for _ in range(self.n_countries)],
+            **kwargs
+        )
```
spjuhel: The base idea would be to do something like:

```python
class PoolOptimizationFixedNumber():
    def __init__(self, nominals, data, bools, alpha, N, fun, **kwargs):
        self.data_arr = data
        self.bools = bools
        self.alpha = alpha
        self.N = N
        self.fun = fun
        self.nominals = np.array(nominals)
        self.n_countries = len(nominals)
        self._pyomoo_problem = ElementwiseProblem(
            n_var=self.data_arr.shape[1],
            n_obj=1,  
            n_constr = 1,
            xl=0,                  
            xu=self.N - 1,
            type_var=int,
            vars=[Integer((0, self.n_countries - 1)) for _ in range(self.n_countries)],
            **kwargs
        )
```

Then in all the places where you use pyomoos `ElementwiseProblem` methods or attribute, you have to change from `<instance_of_PoolOptimisationFixedNumber>.<method|attribute>` to `<instance_of_PoolOptimisationFixedNumber>._pyomoo_problem.<method|attribute>`.

Notes:

- `_pyomoo_problem` is a placeholder name, you should think of a better one.
- I assume `N` is the number of pools? I would rename it `n_pools`.
- You also might want to delegate the init of the pyomoo_problem to a method:

```python
class PoolOptimizationFixedNumber():
    def __init__(self, nominals, data, bools, alpha, N, fun, **kwargs):
        self.data_arr = data
        self.bools = bools
        self.alpha = alpha
        self.N = N
        self.fun = fun
        self.nominals = np.array(nominals)
        self.n_countries = len(nominals)
        self._pyomoo_problem = self._init_optimisation_problem(self.data_arr.shape[1], len(nominals), N, **kwargs)

    @classmethod
    def _init_optimisation_problem(cls, n_var, n_countries, n_pools, **kwargs): # I would also rename n_var to what they are in your case
         return ElementwiseProblem(
            n_var=n_var, 
            n_obj=1,  
            n_constr = 1,
            xl=0,                  
            xu=n_pools - 1,
            type_var=int,
            vars=[Integer((0, .n_countries - 1)) for _ in range(n_countries)],
            **kwargs
        )
```

climada_petals/engine/cat_bonds/pooling_functions.py
spjuhel: See my other comment. Tell me if this is not clear enough!

KaiOBerg: Thanks, Samuel, that helped a lot! I think I solved it :)

chahank:
Note: the tutorial would benefit from reforumalting the making of the TC sets. Now it looks like one needs to take a specific cyclone and perturb it. I would hide that behind a test file.

Otherwise the tutorial runs through well for me.

chahank:
~~I think `pymoo` is mising as a requirement. Any opinion how to handle this @spjuhel ? This is why the tests are failing.~~

Just added it to the .toml file.

chahank:
We should update the `Changelog`.

chahank:
Ok, one test in this PR is failing:

```
    def test_calibration_converges_to_expected_thresholds(self):
        """
        Test that the optimization converges and reports a thresholds for all subareas.
        """
        subareas_calc_dummy = subarea_calculations.SubareaCalculations(
            subareas=self.mock_subarea_calc_test, index_stat=50, initial_guess=(2, 4)
        )
    
        attachment = 0
    
        results, opt_min, opt_max = subareas_calc_dummy._calibrate_payout_fcts(
            haz_int=self.mock_subarea_calc_test.haz_int_dict,
            principal=self.mock_subarea_calc_test.principal,
            attachment=attachment,
            imp_subarea_evt=self.mock_subarea_calc_test.imp_subarea
        )
    
        # optimizer should converge near
>       assert np.allclose(opt_min, [1, 2], atol=1)
E       assert False
E        +  where False = <function allclose at 0x7fb2d5886530>(array([3., 3.]), [1, 2], atol=1)
E        +    where <function allclose at 0x7fb2d5886530> = np.allclose
```

chahank:
I think the tutorial update goes in the very good direction. However, it still looks a bit misleading to me. It looks like computing tc tracks is required to do cat bonds. Can you replace the hazard / exposure parts with a API test file instead? For this, simply export the hazard file that is used currently per country, and then send it to us. We can then upload it to the API.

The reason is that now it reads "create hazard and exposure for Jamaica with Tropical Cyclone Melissa" which suggests that one has to make CatBonds for a specific event. 

The util function for creating the datasets are still very useful, as we have a repository with code to reproduce API files.

chahank:
Concerning the wildfire example: this can be referred to in the Paper to emphasize that this is not about TCs.

chahank:
I just tested the tutorial with another hazard than TC, and it works like a charm. Great job! I would then propose to do one last update of the tutorial and make it hazard "agnostic". So, instead of using variable names like `tc_jamaica`, use `haz_jamaica`. Then, only one more cell needs to be updated:

```
tc_jamaica, exp_jamaica = create_tc_hazards_for_country(country_code=countries[1], storm_id="2025291N11319", bbox=(17.5, 18.75, -78.5, -76)) # create hazard and exposure for Jamaica with Tropical Cyclone Melissa
tc_kit, exp_kit = create_tc_hazards_for_country(country_code=countries[0], storm_id="2017242N16333", bbox=(17.0, 17.5, -62.9, -62.5)) # create hazard and exposure for St. Kitts and Nevis with Tropical Cyclone Irma
tc_belize, exp_belize = create_tc_hazards_for_country(country_code=countries[2], storm_id="2000273N16277", bbox=(15.5, 18.75, -89.5, -87.5)) # create hazard and exposure for Belize with Tropical Cyclone Keith
```

to something like 

```
# Here we demonstrate the module using demonstration data from the climada data API.
# The hazard, exposures and impact function sets should be replaced by target data for your study.
# PLease see the hazard, exposures and impact function tutorials for data loading etc.
from climada.util.api_client import Client
client = Client()
exp_jamaica = client.get_litpop('JAM')
haz_jamaica = client.get_hazard('tropical_cyclone', properties={'country_iso3alpaha'='JAM', 'climate_scenario':'None'}
```

Maybe you can add a section at the bottom of the tutorial where you run all the cells, without the extra explnations, for another hazard - exposures - impact function set.

```
Example for another hazard, exposures, impact function combination: Eartquakes

...
```

This can then be mentioned in the reference paper as a demonstration of the use of the module for any type / resolution of hazard and exposure.

KaiOBerg:
@chahank and @spjuhel 

I made several updates to the module: 

**In the tutorial:** 

1. Replace synthetic tc hazard tracks with obsreved tcs from the API. This works fine and the tracks are already on the API so we dont have to upload another data set.
2. Load exposure from data API. 
3. Add an earthquake cat bond for belize using litpop exposure and the earthquake hazard data set from the API. Impact function is a dummy which currently lives in the create_hazard.py. I guess that should be changed (at least the name of the script) but I wasnt sure if you want to load it to the API or anything? 

**CAT bond simulation**
1. Add a start year variable to function and implement non-overlapping terms. I think this makes the simulation more intuitive. So know you can say, my hazard data covers years x to z, so my simulation should start in x and cover a certain number of terms. 

**Tests**
1. I fixed the failed tests in the premium calculation class. The reason was that I read an excel which requires a certain package to be of a certain version which is not supported in our python 3.12 environment, I replaced the excel with a csv which can be imported without problems. 
2. Subarea calcualtions: The test for the payout function calibration failed because the optimization was not stable enough with only two hazard data points. I now have 10 hazards for calibration which works fine. The problem is that there has to be sufficient data to calibrate the payout functions as there is no fallback. Maybe we can implement a fallback which somehow creates a very rough version of a payout function to still be able to simulate the bond?

The tests still fail, however, I dont think that they are related to my scripts so I am not sure what to do about it...

chahank:
> @chahank and @spjuhel
> 
> I made several updates to the module:
> 
> **In the tutorial:**
> 
>     1. Replace synthetic tc hazard tracks with obsreved tcs from the API. This works fine and the tracks are already on the API so we dont have to upload another data set.

Great. Any reason to use observed instead of probabilistic (synthetic) tracks? 

> 
>     2. Load exposure from data API.
> 
>     3. Add an earthquake cat bond for belize using litpop exposure and the earthquake hazard data set from the API. Impact function is a dummy which currently lives in the create_hazard.py. I guess that should be changed (at least the name of the script) but I wasnt sure if you want to load it to the API or anything?
> 

Exellent.

> 
> **CAT bond simulation**
> 
>     1. Add a start year variable to function and implement non-overlapping terms. I think this makes the simulation more intuitive. So know you can say, my hazard data covers years x to z, so my simulation should start in x and cover a certain number of terms.
> 

Very good for usability.

> 
> **Tests**
> 
>     1. I fixed the failed tests in the premium calculation class. The reason was that I read an excel which requires a certain package to be of a certain version which is not supported in our python 3.12 environment, I replaced the excel with a csv which can be imported without problems.
> 
>     2. Subarea calcualtions: The test for the payout function calibration failed because the optimization was not stable enough with only two hazard data points. I now have 10 hazards for calibration which works fine. The problem is that there has to be sufficient data to calibrate the payout functions as there is no fallback. Maybe we can implement a fallback which somehow creates a very rough version of a payout function to still be able to simulate the bond?

That could be an idea. How complicated would it be?

> 
> 
> The tests still fail, however, I dont think that they are related to my scripts so I am not sure what to do about it...

chahank:
> @chahank and @spjuhel
> 
> I made several updates to the module:
> 
> **In the tutorial:**
> 
>     1. Replace synthetic tc hazard tracks with obsreved tcs from the API. This works fine and the tracks are already on the API so we dont have to upload another data set.

Great. Any reason to use observed instead of probabilistic (synthetic) tracks? 

Also, I noted that the impact functions where not assigned to the exposures. That should be done to use the correct impact function per country. The calibrated set contains I think 9 different ones for different world regions.

> 
>     2. Load exposure from data API.
> 
>     3. Add an earthquake cat bond for belize using litpop exposure and the earthquake hazard data set from the API. Impact function is a dummy which currently lives in the create_hazard.py. I guess that should be changed (at least the name of the script) but I wasnt sure if you want to load it to the API or anything?
> 

Exellent.

> 
> **CAT bond simulation**
> 
>     1. Add a start year variable to function and implement non-overlapping terms. I think this makes the simulation more intuitive. So know you can say, my hazard data covers years x to z, so my simulation should start in x and cover a certain number of terms.
> 

Very good for usability.

> 
> **Tests**
> 
>     1. I fixed the failed tests in the premium calculation class. The reason was that I read an excel which requires a certain package to be of a certain version which is not supported in our python 3.12 environment, I replaced the excel with a csv which can be imported without problems.
> 
>     2. Subarea calcualtions: The test for the payout function calibration failed because the optimization was not stable enough with only two hazard data points. I now have 10 hazards for calibration which works fine. The problem is that there has to be sufficient data to calibrate the payout functions as there is no fallback. Maybe we can implement a fallback which somehow creates a very rough version of a payout function to still be able to simulate the bond?

That could be an idea. How complicated would it be?

> 
> 
> The tests still fail, however, I dont think that they are related to my scripts so I am not sure what to do about it...

KaiOBerg:
> Great. Any reason to use observed instead of probabilistic (synthetic) tracks?

because there are less tracks so it computes faster. Do you think we should change it?

> Also, I noted that the impact functions where not assigned to the exposures. That should be done to use the correct impact function per country. The calibrated set contains I think 9 different ones for different world regions.

true, this was orginally in the wrapper function. I assigned the correct impact functions with the last commit directly in the notebook.

> That could be an idea. How complicated would it be?

I dont think it should take to much time. However, I guess it will result in a very rough payout function with a large basis risk. I could add that and print a warning that the hazard data set is too small/bad.

chahank:
> > Great. Any reason to use observed instead of probabilistic (synthetic) tracks?
> 
> because there are less tracks so it computes faster. Do you think we should change it?

I understand the logic. It depends a bit how much longer the probabilistic sets take to compute. If not too much, I would prefer these as otherwise the tutorial might suggest that one should use historical data only.

> 
> > Also, I noted that the impact functions where not assigned to the exposures. That should be done to use the correct impact function per country. The calibrated set contains I think 9 different ones for different world regions.
> 
> true, this was orginally in the wrapper function. I assigned the correct impact functions with the last commit directly in the notebook.
> 
> > That could be an idea. How complicated would it be?
> 
> I dont think it should take to much time. However, I guess it will result in a very rough payout function with a large basis risk. I could add that and print a warning that the hazard data set is too small/bad.

Thanks for adding this.

climada_petals/engine/cat_bonds/subarea_calculations.py
```diff
@@ -292,7 +326,13 @@ def _calibrate_payout_fcts(self, haz_int, principal, attachment, imp_subarea_evt
                 opt_min, opt_max = result.x
                 subarea_specific_results[subarea] = (opt_min, opt_max)
             else:
-                print(f"Optimization failed for subarea {subarea}: {result.message}")
+                LOGGER.warning(
+                    "Optimization failed for subarea %s: %s. Using fallback thresholds. Basis risk may be higher for this subarea.",
```
chahank: Can the message concerning the fallback thresholds be a bit clearer? I think now if would be very hard for a user to a) understand how the fallback is obtained b) decide whether they should change something in their setup or not.

chahank (approved):
Great, just a small user friendliness update and we can merge.

climada_petals/engine/cat_bonds/subarea_calculations.py
KaiOBerg: I adjusted the message. It now says:
`LOGGER.warning(
                    "Optimization failed for subarea %s: %s. Using fallback thresholds which sets the minimum and maximum parametric index thresholds to the median and maximum parametric index values using all available hazard data. Basis risk may be higher for this subarea. The available hazard data may be insufficient or inadequate to optimize the payout function. Consider adjusting the initial guess or reviewing the hazard intensity data for this subarea.",
                    subarea, result.message,
                )`

KaiOBerg:
I have done several improvements which mainly target the efficient calculation of the parametric index as well as the dataframe containing payout and damage per event (by using vecotrization and batch processing). Before they were both calculated with nested loops which took a lot of time. Now, the notebook runs much faster and I changed the hazard data sets for TCs to the respective STORM data sets. To run the notebook now takes ~10 minutes.

To use the STORM datasets I also created a helper function which reassigns the dates of the events in the STORM data sets, as they are originally all assigned to the year 1980 even though the data set should cover 1000 years of TC tack data. For the CAT bond simulation this is not useful as then all hazard events are attributed to one single CAT bond term. 
I also adjusted the accompanying documentation in the notebook to inform the user about this. 

The helper function is stored in the newly created script "helper_tutorial_cat_bonds.py" which also contains the function used to create the mock impact function for earthquakes. I also added very basic unit tests in a new script for those.

chahank:
> I have done several improvements which mainly target the efficient calculation of the parametric index as well as the dataframe containing payout and damage per event (by using vecotrization and batch processing). Before they were both calculated with nested loops which took a lot of time. Now, the notebook runs much faster and I changed the hazard data sets for TCs to the respective STORM data sets. To run the notebook now takes ~10 minutes.

Excellent! Although, I think ~10minutes for the notebook is a bit long as this is compiled for th documentation. A subset of the storms might do the job?

> 
> To use the STORM datasets I also created a helper function which reassigns the dates of the events in the STORM data sets, as they are originally all assigned to the year 1980 even though the data set should cover 1000 years of TC tack data. For the CAT bond simulation this is not useful as then all hazard events are attributed to one single CAT bond term. I also adjusted the accompanying documentation in the notebook to inform the user about this.

I think this is a central point. Is this also clear in the docstrings where the Hazard set is given by the user? 

Annex question: should we maybe advice to sample a yearset from the hazard if it is not already one? Or how were the dates assigned to the STORM data set?

> 
> The helper function is stored in the newly created script "helper_tutorial_cat_bonds.py" which also contains the function used to create the mock impact function for earthquakes. I also added very basic unit tests in a new script for those.

spjuhel:
> Excellent! Although, I think ~10minutes for the notebook is a bit long as this is compiled for th documentation. A subset of the storms might do the job?

The notebooks are **not** run for the documentation, they are used statically (so outputs are the last generated outputs)

Sorry I haven't put much time in reviewing this lately, but I think the code was pretty good already so far!

chahank:
> > Excellent! Although, I think ~10minutes for the notebook is a bit long as this is compiled for th documentation. A subset of the storms might do the job?
> 
> The notebooks are **not** run for the documentation, they are used statically (so outputs are the last generated outputs)

True, then it should be good. I would then propose to merge unless @KaiOBerg you have further changes planned.

> 
> Sorry I haven't put much time in reviewing this lately, but I think the code was pretty good already so far!

KaiOBerg:
I made some changes based on Chahans comments and the discussion: 

- improve documentation so that users don't miss that the date of an hazard is relevant for the simulation
- perform a code review with copilot -> no major changes but some of it was really useful and makes the code better 
- based on the review with copilot I added networkx to the dependencies in pyproject.toml
- I checked the tests and none of the tests related to the cat bond module fail

chahank:
Thanks for the update!

One thing I need to double check: why is the networkx package required? We should add new dependencies only when it is absolutely necessary.

KaiOBerg:
@chahank, I replaced networkx and removed it from the dependencies.

chahank:
Excellent, thanks! Please merge then, great work!

PR #192, opened 2026-01-12
Network module update part 1: Network base class and NetworkCalcs

luseverin:
This PR is the first part of a larger PR refactoring the engine's network module. 
Changes proposed in this PR:
- First draft `Network` base class - data store class
- First draft `NetworkCalcs` class - class handling calculations on networks and associated `igraph.graph` objects
- Tutorial showcasing a possible use case

Suggestions for review:
As this is still preliminary work, I would suggest the reviewer to focus on the following points:
- Purpose of the module. Look at the tutorial; Can you understand the purpose of the module and its core functionalities? Do you see any major drawbacks from a usage perspective? Any core functionality missing?
- Structure of the module. Look at the `Network`, `GraphCalcs` and `NetworkCalcs` classes; do you see any issue in terms of how the module is currently stuctured? Would you do things differently?

Some more details:
- The network class is intended as a data storage class primarily
- The nw_calc module contains two classes: `NetworkCalcs` and `GraphCalcs`. `GraphCalcs` handles the graph representation of a network (using igraph) and gathers graph-based functions. It also takes care of conversion between graph and networks. `NetworkCalcs` gathers higher-level wrappers including dependency calculation and failure cascade.
- It might not be possible to run the entire tutorial at the moment as some functionalities are to be added in coming PRs. Please check-out the feature/nw_update branch if you want to run it entirely

network_class_diag_v2

This PR fixes #

### PR Author Checklist

### PR Reviewer Checklist


chahank:
Great! Thank you for separating this larger PR into sub PRs. Please let reviewers know once you are ready for review.

climada_petals/engine/networks/nw_base.py
```diff
@@ -0,0 +1,543 @@
...
+class Network:
+
+    # map methods
+    plot_infra = infra_plot
+    plot_pop = population_plot
+    plot_dep = dep_plot
+    plot_access = access_plot
+
+    def __init__(self,
+                 edges=gpd.GeoDataFrame(),
```
spjuhel: Avoid putting mutable objects as arguments!
Put None as a default value and use:

`edges = gpd.GeoDataFrame() if edges is None else edges`

In the body.

(See for instance: https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python)

climada_petals/engine/networks/nw_base.py
```diff
@@ -0,0 +1,543 @@
...
+        if edges.empty:
+            edges = gpd.GeoDataFrame(
+                columns=['from_id', 'to_id', 'id', 'orig_id', 'geometry'],
+                geometry='geometry', crs='EPSG:4326')
+        if nodes.empty:
+            nodes = gpd.GeoDataFrame(
+                columns=['id', 'orig_id', 'geometry'],
+                geometry='geometry', crs='EPSG:4326')
+
+        if not hasattr(edges, 'orig_id'):
```
spjuhel:
```suggestion
        if 'orig_id' not in edges.columns:
```
Is more standard for dataframes

climada_petals/engine/networks/nw_base.py
```diff
@@ -0,0 +1,543 @@
...
+
+        See Also
+        --------
+        geopandas.GeoDataFrame.to_crs : Underlying reprojection method
+        """
+        self.nodes = self.nodes.to_crs(crs)
+        self.edges = self.edges.to_crs(crs)
+
+    @classmethod
+    def from_nws(cls, networks):
```
spjuhel:
```suggestion
    def from_networks(cls, networks):
```

climada_petals/engine/networks/nw_base.py
```diff
@@ -0,0 +1,543 @@
...
+
+        for net in networks:
+            edge_gdf = net.edges.reset_index(drop=True)
+            node_gdf = net.nodes.reset_index(drop=True)
+            edge_gdf['from_id'] = edge_gdf['from_id'] + id_counter_nodes
+            edge_gdf['to_id'] = edge_gdf['to_id'] + id_counter_nodes
+            node_gdf['id'] = range(id_counter_nodes,
+                                   id_counter_nodes+len(node_gdf))
+            id_counter_nodes += len(node_gdf)
+            edges = pd.concat([edges, edge_gdf])
```
spjuhel: Much more efficient and adapted to initiate empty lists before the for loop, populate and then call concat once on the list after the for loop. Also you then don't need to create empty dataframe before.
See suggestion

climada_petals/engine/networks/nw_base.py
```diff
@@ -0,0 +1,543 @@
...
+            nodes = pd.concat([nodes, node_gdf])
```
spjuhel:
```suggestion
        id_counter_nodes = 0
        edges = []
        nodes = []
        for net in networks:
            edge_gdf = net.edges.reset_index(drop=True)
            node_gdf = net.nodes.reset_index(drop=True)
            edge_gdf['from_id'] = edge_gdf['from_id'] + id_counter_nodes
            edge_gdf['to_id'] = edge_gdf['to_id'] + id_counter_nodes
            node_gdf['id'] = range(id_counter_nodes,
                                   id_counter_nodes+len(node_gdf))
            id_counter_nodes += len(node_gdf)
            edges.append(edge_gdf)
            nodes.append(node_gdf)
        edges = pd.concat(edges)
        nodes = pd.concat(nodes)
```

climada_petals/engine/networks/nw_base.py
```diff
@@ -0,0 +1,543 @@
...
+        geopandas.read_feather : Underlying deserialization method
+        """
+        path_load = Path(path_load)
+        zip_path = path_load / f"{savename}.zip"
+
+        nodes = gpd.GeoDataFrame()
+        edges = gpd.GeoDataFrame()
+
+        if not zip_path.exists():
+            print(f"Archive {zip_path} not found")
```
spjuhel: Prefer LOGGER.info()

```suggestion
            LOGGER.info("Archive %s not found", zip_path)
```

climada_petals/engine/networks/nw_base.py
```diff
@@ -0,0 +1,543 @@
...
+                print(f"Nodes file {nodes_name} not found in archive")
+
+            if edges_name in zf.namelist():
+                with zf.open(edges_name) as f:
+                    edges = gpd.read_feather(io.BytesIO(f.read()))
+            else:
+                print(f"Edges file {edges_name} not found in archive")
+
+        return cls(edges=edges, nodes=nodes)
+    def update_network_from_graphs(self, graphs):
```
spjuhel: This method name is misleading, it does not update the Network object, but overwrite both its edges and nodes.

Consider maybe:

```suggestion
    @classmethod
    def from_graphs(cls, graphs):
    
    [...]
    
        return cls(nodes=nodes, edges=edges)
    
```

climada_petals/engine/networks/nw_base.py
```diff
@@ -0,0 +1,543 @@
...
+        >>> graph = network.to_graph(directed=True)
+        >>> graph.vcount()  # Number of vertices
+        >>> graph.ecount()  # Number of edges
+
+        See Also
+        --------
+        update_network_from_graphs : Synchronize Network from Graph
+        igraph.Graph.DataFrame : Underlying graph construction method
+        """
+        self.directed = directed
```
spjuhel: Avoid creating new attributes to the object outside the `__init__`.
Move directed `directed` to be an argument of `_from_vs`, `_from_es`

climada_petals/engine/networks/nw_base.py
```diff
@@ -0,0 +1,543 @@
...
+        if not self.edges.empty:
+            graph = self._from_es(
+                gdf_edges=self.edges, gdf_nodes=self.nodes)
+        else:
+            graph = self._from_vs(
+                gdf_nodes=self.nodes)
+        return graph
+    def _remove_namecol(self, gdf_nodes):
```
spjuhel, line 418: At least one blank line between methods (consider using black/ruff!)

```suggestion

    def _remove_namecol(self, gdf_nodes):
```

climada_petals/engine/networks/nw_base.py
```diff
@@ -0,0 +1,543 @@
...
+        gdf_nodes : gpd.GeoDataFrame or None
+            GeoDataFrame potentially containing a 'name' column
+
+        Returns
+        -------
+        gpd.GeoDataFrame or None
+            GeoDataFrame with 'name' column removed if it existed, or None
+        """
+        if gdf_nodes is not None:
+            if hasattr(gdf_nodes, 'name'):
```
spjuhel:
```suggestion
            if 'name' in gdf_nodes.columns:
```

climada_petals/engine/networks/nw_base.py
spjuhel: Also is there any reason:
- to create empty networks?
- to have empty nodes but not empty edges? (and vice-versa)

`edges` is very misleading for someone that works with graphs, as one would expect each edge to be a connection between two nodes. I would rename to `lines` maybe?

climada_petals/engine/networks/nw_base.py
```diff
@@ -0,0 +1,543 @@
...
+            Graph with n vertices and 0 edges, where n = len(gdf_nodes)
+        """
+        gdf_nodes = self._remove_namecol(gdf_nodes)
+        vertex_attrs = gdf_nodes.to_dict('list')
+        return ig.Graph(
+            n=len(gdf_nodes),
+            vertex_attrs=vertex_attrs,
+            directed=self.directed)
+
+    def initialize_funcstates(self):
```
spjuhel, line 510: This method will fail if nodes or edges is empty/None.

As discussed, is there any reason to not do this at init time, rather than leaving the user do it?

In this case, you might consider keeping the method and call it in init to allow "resetting" an object.

climada_petals/engine/networks/nw_base.py
```diff
@@ -0,0 +1,543 @@
...
+        >>> network.initialize_funcstates()
+        >>> network.nodes['func_tot']  # All values are 1
+        >>> network.edges['imp_dir']   # All values are 0
+        """
+        self.edges[['func_internal','func_tot']] = 1
+        self.nodes[['func_internal','func_tot']] = 1
+        self.edges['imp_dir'] = 0
+        self.nodes['imp_dir'] = 0
+
+    def initialize_capacity(self, dep_table):
```
spjuhel, line 537: Consider adding dep_table to `__init__` as an optional parameter, and call that function if it is not `None`.

climada_petals/engine/networks/nw_base.py
```diff
@@ -0,0 +1,543 @@
...
+        initialize_supply : Initialize supply attributes for enduser dependencies
+        """
+        for __, row in dep_table.iterrows():
+            source = row['source']
+            target = row['target']
+            self.nodes[f'capacity_{source}_{target}'] = 0
+            self.nodes.loc[self.nodes['ci_type']==f'{source}',f'capacity_{source}_{target}'] = 1
+            self.nodes.loc[self.nodes['ci_type']==f'{target}',f'capacity_{source}_{target}'] = -1
+
+    def initialize_supply(self, dep_table):
```
spjuhel, line 581: Is there a reason to `initialize_capacity` but not `initialize_supply`,
If not, consider having a public method `initialize_dependencies` that call both methods (and make them private)

spjuhel (changes requested):
Ok, first part of the review on the `nw_base.py` file.
See my comments.

climada_petals/engine/networks/nw_calcs.py
```diff
@@ -0,0 +1,1526 @@
...
+        directed : bool, optional
+            Whether to build a directed igraph representation. Default is ``True``.
+        friction_surf : object, optional
+            Friction surface used for duration-based linking. Default is ``None``.
+
+        Notes
+        -----
+        The graph is lazily built and cached on first access via `graph`.
+        """
+        self.parent = parent #parent nw calc object
```
spjuhel: Consider changing to `network_calc`. `parent` suggests child-parent class relationship, which is not the case.

climada_petals/engine/networks/nw_calcs.py
```diff
@@ -0,0 +1,1526 @@
...
+
+        Notes
+        -----
+        The method warns if node/edge columns are not ordered as expected by igraph.
+        """
+        if self.network.nodes.columns[0] != 'id' or \
+              self.network.edges.columns[0] != 'from_id' or self.network.edges.columns[1] != 'to_id':
+            LOGGER.warning("Network nodes or edges columns are not properly ordered for graph generation." \
+            "igraph expects 'id' in position 0 for nodes and 'from_id', 'to_id' in position 0 and 1 respectively for edges."
+            "Please reorder the columns of your nodes and edges using ordered_network() function.")
```
spjuhel: This looks unclear. 

Is this a requirement from `igraph.Graph.DataFrame()`? 
This check should be probably be done in the `to_graph` method of `Network`.
Also, it should probably check for the presence of the column and reorder by itself. Relying on correct column order in a DataFrame should be easy to avoid.

climada_petals/engine/networks/nw_calcs.py
```diff
@@ -0,0 +1,1526 @@
...
+        self._graph = self.network.to_graph(directed=self.directed)
+        return self._graph
+
+    @property
+    def graph(self):
+        if self._graph is None:
+            return self.build_graph()
+        return self._graph
+
+    def invalidate(self):
```
spjuhel:
```suggestion
    def reset_graph_edges(self):
```
Would be more explicit

climada_petals/engine/networks/nw_calcs.py
```diff
@@ -0,0 +1,1526 @@
...
+        Notes
+        -----
+        Distances are approximated by converting meters to degrees using
+        :data:`~climada.util.constants.ONE_LAT_KM`.
+        """
+
+        gdf_vs = self.graph.get_vertex_dataframe()
+        #if ci_type is not None:#filter to ci_type
+        #    gdf_vs = gdf_vs[gdf_vs.ci_type==ci_type]
+        # Use 'weak' mode for directed graphs to treat them as undirected for connectivity
```
spjuhel: It looks like you have a special treatment for directed graphs. I would explain it in the doc.

I am also unsure whether weak components are really what you want. If the directed graph does represent one-way links, connecting nodes assuming the links can be used both way might be plain wrong.

I would suggest setting "mode" via an optional argument with an explicit name like `directed_graph_connextivity` and reflect on what default value (or no default value at all and forcing the user to make a choice)

climada_petals/engine/networks/nw_calcs.py
```diff
@@ -0,0 +1,1526 @@
...
+            try:
+                dists, ix_match = _ckdnearest(
+                    gdf_a, gdf_b, dist_thresh=dist_thresh)
+                source = gdf_a.iloc[np.where(dists == min(dists))[
+                    0]].index[0]
+                target = gdf_b.loc[ix_match[np.where(dists == min(dists))[
+                    0]]].index[0]
+                v_ids_source.append(source)
+                v_ids_target.append(target)
+            except (IndexError, KeyError):
```
spjuhel: Consider logging absence of match in debug / info

climada_petals/engine/networks/nw_calcs.py
```diff
@@ -0,0 +1,1526 @@
...
+        target_attrs : dict
+            Vertex attribute filters for target candidates.
+        link_attrs : dict, optional
+            Edge attributes for created links.
+        dist_thresh : float, optional
+            Maximum distance (in meters) to allow links. Default is ``np.inf``.
+        bidir : bool, optional
+            If ``True``, add reverse links as well. Default is ``False``.
+        k : int, optional
+            Number of nearest sources per target. Default is ``5``.
```
spjuhel: Why 5?
Consider not putting a default value. This is a choice that should be carefully considered by the user.

climada_petals/engine/networks/nw_calcs.py
```diff
@@ -0,0 +1,1526 @@
...
+        Parameters
+        ----------
+        source_attrs : dict
+            Vertex attribute filters for source candidates.
+        target_attrs : dict
+            Vertex attribute filters for target candidates.
+        link_attrs : dict, optional
+            Edge attributes for created links.
+        dist_thresh : float, optional
+            Maximum distance (in meters) to allow links. Default is ``np.inf``.
```
spjuhel: `np.inf` looks like a dangerous default value.

Consider not putting a default value.

climada_petals/engine/networks/nw_calcs.py
```diff
@@ -0,0 +1,1526 @@
...
+        ----------
+        source_attrs : dict
+            Vertex attribute filters for source candidates.
+        target_attrs : dict
+            Vertex attribute filters for target candidates.
+        via_attrs : dict
+            Edge attribute filters for allowable paths.
+        link_attrs : dict
+            Edge attributes for created dependency links.
+        dist_thresh : float, optional
```
spjuhel: See other comments on default values.

climada_petals/engine/networks/nw_calcs.py
```diff
@@ -0,0 +1,1526 @@
...
+        subgraph_graph_vsdict = self._get_subgraph2graph_vsdict(self.graph, subgraph)
+
+        # Vectorize the dictionary lookup
+        subgraph_ids = np.array(sorted(subgraph_graph_vsdict.keys()))
+        graph_ids = np.array([subgraph_graph_vsdict[k] for k in subgraph_ids])
+
+        # select only those for which specified attrs apply
+        cache = {}
+        df_vs_target = GraphCalcs._filter_vertices(subgraph, target_attrs, cache)
+        df_vs_source = GraphCalcs._filter_vertices(subgraph, source_attrs, cache)
```
spjuhel: Are the vertices bound to change within a `GraphCalcs` instance?
If not, consider having the cache as a self._cache attribute instead and handling the cache at the instance level rather that via obscure methods. (consider looking into `lru_cache`)

See for instance: https://stackoverflow.com/questions/60207688/hashing-a-pandas-dataframe-for-calculated-column-caching

Note, caching adds a layer of complexity for testing and is very error-prone. Carefully consider if the avoided computation time is really worth it, or if a simple computation of these dataframes at init and attribute creation isn't more appropriate.

climada_petals/engine/networks/nw_calcs.py
```diff
@@ -0,0 +1,1526 @@
...
+        else:
+            target_indices = np.where(is_target)[0]
+            for idx_in_all in target_indices:
+                actual_idx = all_ids[idx_in_all]
+                orig_func = self.graph.vs[int(actual_idx)]['func_tot']
+                self.graph.vs[int(actual_idx)]['func_tot'] = min(capa_suff[idx_in_all], orig_func)
+
+        #delete large objects to avoid memory issues
+        del capa_rec, func_capa, capa_suff, adj_sub, func_thresh
+        gc.collect()
```
spjuhel: local variables are always deleted at the end of a function. This code should not be needed.

climada_petals/engine/networks/test/test_networks.py
```diff
@@ -0,0 +1,2245 @@
...
+    def test_funcstates_sum(self, graph_calcs):
+        """Test summing functional states"""
+        graph_calcs.build_graph()
+
+        v_sum, e_sum = graph_calcs._funcstates_sum()
+
+        assert isinstance(v_sum, (int, float))
+        assert isinstance(e_sum, (int, float))
+        assert v_sum > 0
+        assert e_sum > 0
```
spjuhel: Check the actual value expected

climada_petals/engine/networks/test/test_networks.py
```diff
@@ -0,0 +1,2245 @@
...
+        graph_calcs_with_remote_node_missing_edge.build_graph()
+        initial_edge_count = graph_calcs_with_remote_node_missing_edge.graph.ecount()
+        graph_calcs_with_remote_node_missing_edge.link_clusters(
+            dist_thresh=np.inf,
+            link_attrs={'ci_type': 'cluster_link'},
+        )
+
+        # Verify method completes
+        assert graph_calcs_with_remote_node_missing_edge.graph.ecount() == initial_edge_count + 1
+        assert 'cluster_link' in graph_calcs_with_remote_node_missing_edge.graph.es['ci_type']
```
spjuhel: Can you check that the added edge does link the correct nodes?

climada_petals/engine/networks/test/test_networks.py
```diff
@@ -0,0 +1,2245 @@
...
+        graph_calcs.link_vertices_closest_k(
+            source_attrs={'ci_type': 'road'},
+            target_attrs={'ci_type': 'people'},
+            link_attrs={'ci_type': 'link_road_people'},
+            dist_thresh=1000,
+            bidir=False,
+            k=1
+        )
+
+        # Should add at least one edge
```
spjuhel:
```suggestion
        # Should not add any edge
```
?

climada_petals/engine/networks/test/test_networks.py
spjuhel, line 1: This file is really long, consider splitting it for the different classes tested.

climada_petals/engine/networks/nw_calcs.py
spjuhel, line 1: Likewise, this file is also quite long. Not critical though. But consider putting `NetworkCalc` in a separate file for instance.

climada_petals/engine/networks/nw_calcs.py
```diff
@@ -0,0 +1,1526 @@
...
+"""
+import logging
+import igraph as ig
+import numpy as np
+import pandas as pd
+import geopandas as gpd
+import pyproj
+from tqdm import tqdm
+import timeit
+import gc
```
spjuhel: Several imports are unused (there should be an autofix for this in your IDE)

climada_petals/engine/networks/nw_calcs.py
```diff
@@ -0,0 +1,1526 @@
...
+
+            edges_dys = self.graph.es.select(ci_type='power_line'
+                                             ).select(func_tot_eq=0)
+            reverse_edges = [(edge.target, edge.source) for edge in edges_dys]
+            eids = self.graph.get_eids(pairs=reverse_edges, path=None,
+                                       directed=True, error=True)
+            self.graph.es[eids]['func_tot'] = 0
+            LOGGER.info(f"""Using updated power line algorithm: dysfunc edges before:
+                  {len(edges_dys)}, after: {len(self.graph.es.select(ci_type='power_line'
+                                             ).select(func_tot_eq=0))}""")
```
spjuhel: Use lazy % formatting in logging functions

(See https://stackoverflow.com/questions/66993387/use-lazy-formatting-in-logging-functions-pylint-error-message)

climada_petals/engine/networks/nw_calcs.py
```diff
@@ -0,0 +1,1526 @@
...
+            self.powercap_from_clusters(p_source=p_source, p_sink=p_sink,
+                                        demand_ci='people', source_var=source_var, demand_var=demand_var)
```
spjuhel: This raises an error that this method does not exist in my linter.

climada_petals/engine/networks/nw_calcs.py
```diff
@@ -0,0 +1,1526 @@
...
+                - Resets network IDs to account for newly created edges
+                - Invalidates cached graph data after completion
+        """
+        delta = -1
+        cycles = 0
+        while delta != 0:
+            LOGGER.info(
+                f'Updating functional states. Current delta: {delta}')
+            func_states_vs, func_states_es = self.graph_calc._funcstates_sum()
+            self.graph_calc._update_internal_dependencies(
```
spjuhel: Not critical, but it is not considered good practice to access "protected" members (prefixed by `_`) outside the parent class of the member.

I think what you want here is to have graph_calc as a protected member of `NetworkCalcs`, and `_update_internal_dependencies` as a non-protected member of `GraphCalcs`.

The same applies to other protected methods called below.

spjuhel (changes requested):
OK, I have finished a first pass on the whole code.

The code is a bit complex and dense, but I think that's expected for such a complex module, and also acceptable for petals. The overall structure looks decent enough, and I don't see any major issues.

I think however that the "user interface" would benefit from a more extensive documentation in the `NetworkCalcs` class. 

Some additional things that would be nice:
- Blackening the code (rather after all comments have been addressed, as otherwise the line references are going to be lost).
- type hints in the user-interface methods.

Other comments on technical aspects are directly in the files. Some suggestions are mentioned only once but should be applied everywhere relevant.

(Good work!! It's probably one of the biggest and most interesting contribution to CLIMADA)

climada_petals/engine/networks/nw_base.py
luseverin: Good point! I think it makes sense to have nodes without edges (think for instance of a network with only point exposures). Regarding a network with edges but no nodes, I am less sure if it makes sense.. I think you could start with a lines-only network and use some of the preprocessing functions to add the  edges' endpoints as nodes (see add_endpoints function https://github.com/CLIMADA-project/climada_petals/blob/feature/nw_update/climada_petals/engine/networks/nw_preps.py#L213C5-L213C18, which requires a network object as input).

luseverin: Regarding naming, would you then also suggest to rename nodes to points, to be consistent?

climada_petals/engine/networks/nw_base.py
luseverin: Yes that makes sense, I will change that for clarity

luseverin: Is this then bad practice to have the network attribute being changed "silently" within a same instance of NetworkCalcs ?

climada_petals/engine/networks/nw_base.py
luseverin: I think I do not call these methods as default at init because they are only required for later propagation and cascades. For the preprocessing steps they are not yet required. Furthermore, the initialize_capacity and supply requires the dependency table, but the dependency table is not necessary for the rest of the preprocessing. Also, I wanted to leave the freedom to start with not fully functional network so that's why I did not call it by default at init.

What I could do is add initialize_supply and initialize_funcststates as args in the init of NetworkCalcs, initialize_capacity needing to be triggered anyways.

climada_petals/engine/networks/nw_base.py
luseverin: One could come up with a custom base state for the supply and still wants to initialize the capacity, I think.

climada_petals/engine/networks/nw_base.py
luseverin: I think I would rather move these to utils, and then call them at init in NetworkCalcs, does that sound reasonable?

climada_petals/engine/networks/nw_calcs.py
luseverin: Yes, this is a (weird) requirement from igraph. I will move these checks and reformatting to `to_graph`, as suggested.

climada_petals/engine/networks/nw_calcs.py
luseverin: I think you are right. I am still a bit unclear about the use of directed/undirected graphs in the model. For roads, the graph needs to be undirected (or it must be made bidirectional), in order to allow travel along an edge in both direction. However, we require a directed graph for propagation to ensure propagation is made from source to target. I think other CIs such as power infrastructure might also require directed graph.
Anyways, thanks for raising this :)

climada_petals/engine/networks/nw_calcs.py
luseverin: You are right I'll remove the default values

climada_petals/engine/networks/nw_calcs.py
luseverin: Thanks for the suggestions. I removed some of the caching where it was not necessary. I will rethink about it when I'll do some code optimization later.

climada_petals/engine/networks/nw_calcs.py
luseverin: Yes sorry, I remove these functions as I am planning to add them in a later PR. I have not tested or updated any of the power-related functions yet (this is low-priority for the moment)

climada_petals/engine/networks/nw_calcs.py
luseverin: Good point, I protected `graph_calc` and unprotected methods that are meant to be accessed outside of the parent class.

climada_petals/engine/networks/test/test_networks.py
luseverin: Done.

climada_petals/engine/networks/test/test_networks.py
luseverin: Yes! Good catch :)

luseverin:
> OK, I have finished a first pass on the whole code.
> 
> The code is a bit complex and dense, but I think that's expected for such a complex module, and also acceptable for petals. The overall structure looks decent enough, and I don't see any major issues.
> 
> I think however that the "user interface" would benefit from a more extensive documentation in the `NetworkCalcs` class.
> 
> Some additional things that would be nice:
> 
>     * Blackening the code (rather after all comments have been addressed, as otherwise the line references are going to be lost).
> 
>     * type hints in the user-interface methods.
> 
> 
> Other comments on technical aspects are directly in the files. Some suggestions are mentioned only once but should be applied everywhere relevant.
> 
> (Good work!! It's probably one of the biggest and most interesting contribution to CLIMADA)

Thanks a lot @spjuhel for reviewing it so rapidly! I have addressed your comments (all very sensible). Please let me know if there is anything else.
I have written down things that still need to be done from my side or further discussed:
- call initializing functions at init of `Network` or `NetworkCalcs`; I will reflect on what is the most convenient from a user-perspective.
-  Improve documentation; this includes functions' docstrings and more advanced examples in notebook (e.g. show examples of each linking functions and in which cases they can be used)
- Blackening of the code & type hints (both on my todo list)
- Encode variable names in a `constant.py` file to avoid hard-coded values in code

@spjuhel Is there anything else from your side?
If not, I will proceed with these + preparing the next PRs (`nw_preps` and `nw_utils` as well as `nw_impact_calcs`)

PR #201, opened 2026-07-13
Feature/quake

chahank:
Changes proposed in this PR:
- A simple Earthquake model. Based on the previous MatLab version. 

This PR fixes #

### PR Author Checklist

### PR Reviewer Checklist


chahank:
- Tests currently fail to build for reasons not related to this PR. 

- The underlying file currently needs to be downloaded manually (the download link requires an email, thus it is not automated). A separate PR can then take care of uploading the file to the CLIMADA API and make it available to users as fallback option (and as test file).

climada_petals/hazard/earthquake.py
```diff
@@ -0,0 +1,519 @@
...
+    if np.any(dist + corr < 0):
+        raise ValueError("'dist + corr' must not be negative.")
+
+    max_mmi = 1.5 * (mag - 1.0)
+
+    # At zero distance and zero correction, log(0) deliberately produces
+    # -infinity. The resulting positive infinity is subsequently capped at the
+    # magnitude-dependent maximum MMI.
+    with np.errstate(divide="ignore", invalid="ignore"):
+        mmi = (
```
NicolasColombi, line 516: I assume this comes from the paper you cite in the description. Is there a way to add a test using the `mmi` computed independently from this script, for instance from a known case in the paper you use ? This would test that you accidentally did not made a mistake in the formula or parameters and ensure that in the future, if someone would change these the parameters or formula, the test would fail.

If you could find even just one data point with `lat,lon,depth` of the earthquake, and one value of `mmi` at a distance `dist` from the earthquake, this would validate your equation.

climada_petals/hazard/test/test_earthquake.py
```diff
@@ -0,0 +1,276 @@
...
+        self.centroids = Centroids.from_lat_lon(
+            lat=np.array([0.0, 10.0]),
+            lon=np.array([0.0, 10.0]),
+        )
+        self.catalogue = pd.DataFrame(
+            {
+                "lat": [0.0, 10.0, 5.0, 50.0],
+                "lon": [0.0, 10.0, 5.0, 50.0],
+                "mw": [7.0, 2.0, 6.0, 6.0],
+                "depth": [10.0, 10.0, 10.0, 10.0],
```
NicolasColombi: Maybe is not necessary, but why not have different depth values here ?

NicolasColombi (commented):
Thanks for the implementation! :raised_hands: Looks good to me, I think we could add an integration test using some expected values of `mmi` from the literature and test the exact values and structure of the end product.

Please also update the change log and cancel the cell outputs from the jupyter notebook.

climada_petals/hazard/earthquake.py
chahank: This is a good point, but not available directly. The references is an older book with curves, but no example data points. There are also alternative parametrization presented. I thus added a note on this in the module, but could not add an independent validation point.

chahank:
Thank you for the review! I addressed all points (I hope it is satisfactory). I will merge and close, if there is critical issue we can reopen.

PR #204, opened 2026-09-08
perf: sparsify landslide event sampling instead of materialising the …

### LINKED ISSUE #35: LS probabilistic event sampling causes memory issues
When running the LS.from_prob() method, this frequently causes the kernel to die. 
Could be that the implementation of the event sampling function from the probability matrix (https://github.com/CLIMADA-project/climada_petals/blob/1568cfe019342f8be4015db3295f7ceb5aa55a4c/climada_petals/hazard/landslide.py#L40) destroys the sparse matrix, hence massively blowing up memory.

EnvDroneSense:
Changes proposed in this PR:
- `sample_events` sparsifies each sampled year and `vstack`s the rows, instead of passing a list of `n_years` dense rows to `sparse.csr_matrix`.
- `from_prob` sets `fraction.data` directly instead of using a CSR fancy-index assignment.
- Two tests, one bounding peak memory as a regression guard.

This PR fixes #35

`prob_matrix` reaches `sample_events` dense, and each draw returns one dense year. So when `sparse.csr_matrix(events)` runs there are two full copies of the simulation alive: the list of dense rows, and the dense `(n_years, n_pixels)` array scipy builds from it before compressing. Peak is `2 x n_years x n_pixels x 8` bytes however rare landslides are 695 MB at 1 000 years on the packaged raster, for a result holding 13 nonzeros.

`Landslide.from_prob` on the repo's `test_ls_prob.tif` (43 200 px), old and new in one process, memory and time measured separately:

```
                              peak old  peak new      less    t old    t new
                                  (MB)      (MB)               (ms)     (ms)
sample_events, 100 yr             69.1       1.0     69.1x       78       71    1.10x
sample_events, 500 yr            345.7       1.3    266.1x      393      356    1.11x
sample_events, 1000 yr           691.4       1.7    406.9x      785      709    1.11x
from_prob, 500 yr                349.5       5.3     66.1x      407      369    1.10x
from_prob, 1000 yr               695.2       5.5    126.6x      799      727    1.10x
```

The ratio grows with `n_years x n_pixels`; the old peak scales with the problem, the new one with the result.

**Negligible speed up:** it reads ~1.10x here, but in previous tests it has been 0.95x so the impact on speed is mainly noise. But that isn't a problem because the improvement on memory is way larger. The call is dominated by `n_years` scipy `rvs` invocations, which this does not touch, so no time effect is the claim I would defend.

**Output is bit-identical.** `sample_event_from_probs` is called the same number of times, in the same order, with the same arguments; nothing touches the RNG. Verified for both distributions at 100 and 500 years. A variant drawing `k` years per vectorised `rvs` call was measured too, same time and only 11.8x less peak, and rejected because it changes the draw order.

Two notes on the issue. It suggests the sampler *"destroys the sparse matrix"*, but nothing upstream is sparse; the cost is the two dense copies. And #39 modified this function in 2022 without changing that construction, which matches @chahank's doubt at the time that #39 would resolve this.

The memory guard's bound derives from `n_years x n_pixels x 8 bytes` and fails on the current implementation with `AssertionError: 345678407 not less than 17280000.0`. The second test covers `from_prob`'s `fraction` deterministically; the equivalent assertions in `test_from_prob` sit behind `if LS_prob.intensity.size`, which is `nnz` for a sparse matrix, so they are skipped when a draw produces nothing. That is a separate test-hygiene issue.

### PR Author Checklist

- [x] [Documentation](https://climada-python.readthedocs.io/en/latest/guide/Guide_PythonDos-n-Donts.html#2.--Commenting-&-Documenting) updated

### PR Reviewer Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Code freeze date: YYYY-MM-DD
 ### Changed
 
 ### Fixed
+- `Landslide.sample_events` sparsifies each sampled year instead of building the whole simulation densely, reducing peak memory by 60x-120x in `Landslide.from_prob`. Output unchanged. [#35](https://github.com/CLIMADA-project/climada_petals/issues/35)
 
 ### Deprecated
 
--- a/climada_petals/hazard/landslide.py
+++ b/climada_petals/hazard/landslide.py
@@ -75,11 +75,13 @@ def sample_events(prob_matrix, n_years, dist='binom'):
     Landslide.from_prob()
     """
 
-    events = [
-        sample_event_from_probs(prob_matrix, n_samples=1, dist=dist)
+    # Sparsify each year as it is drawn: passing a list of dense rows to
+    # sparse.csr_matrix() holds two copies of the whole simulation at once.
+    rows = [
+        sparse.csr_matrix(sample_event_from_probs(prob_matrix, n_samples=1, dist=dist))
         for i in range(n_years)
         ]
-    return sparse.csr_matrix(events)
+    return sparse.vstack(rows, format="csr")
 
 
 def sample_event_from_probs(prob_matrix, n_samples, dist):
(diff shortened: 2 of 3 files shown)


EnvDroneSense:
The associated benchmarking script:
[pr_benchmark_landslide.py](https://github.com/user-attachments/files/31959698/pr_benchmark_landslide.py)

